PR #789, opened 2021-08-20
Fix Typo1

JoshuaSBrown:
Change 'no' to 'on'.


dvstans:
The change was made to an auto-generated file rather than a source file.

PR #790, opened 2021-08-20
typos-2

JoshuaSBrown:


dvstans:
Need to fix the source file, not generated HTML

PR #826, opened 2022-08-01
Fixes to account for hashed password

JoshuaSBrown:
Store passwords as hashes.


JoshuaSBrown:
This has been confirmed to be working, so merging into master.

PR #845, opened 2022-10-13
A few small fixes

JoshuaSBrown:
Fixes issues with log file path.


JoshuaSBrown:
The generate repo config script was using the wrong variable, was using the web server variable as opposed to the repo variable.

PR #860, opened 2022-11-18
Attempting to create more specific targets in CMake build system

JoshuaSBrown:
Well this was a pain to get sort of working.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,24 +9,19 @@ set(DataFed_VERSION ${DataFed_MAJOR_VERSION}.${DataFed_MINOR_VERSION}.${DataFed_
 set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(THREADS_PREFER_PTHREAD_FLAG ON)
-
+set(CMAKE_SKIP_INTALL_ALL_DEPENDENCY TRUE)
 find_package(Threads REQUIRED)
 
 # Globus version to compile repo server with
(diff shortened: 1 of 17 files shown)


JoshuaSBrown:
The goal of this pr to move as much of the build steps as can be moved into the CMake side of things. The other goal was to better separate installation of "services" from the actual "servers". This pr still needs a lot of work.

Ultimately we should be able to pick and build a specific target without having to build or install any of the other targets.

JoshuaSBrown:
@dvstans I think this addressed what you wanted. Please take a look and make sure. If so I'll clean it up. You should now have the ability to install individual servers, libraries and services using individual targets. 

```bash
cmake --build build --target install-core-server
cmake --build build --target install-core-service
cmake --build build --target install-repo-server
cmake --build build --target install-repo-service
cmake --build build --target install-authz
cmake --build build --target install-web-server
cmake --build build --target install-web-service
```

If there are a few more additional targets you think would be useful to define let me know though I don't want to spend too much more time on this given there are other more pressing priorities.

JoshuaSBrown:
NOTES after meeting the following changes need to be made:

1. Don’t overwrite master config file if already exist
2. If no master config file exists generate them automatically on build – should not require a separate generate_script on install call
3. Package specific config files should be automatically generated from settings in the master config file.

./DataFed/scripts/generate_datafed.sh – run during build
./DataFed/scripts/generate_core_config.sh – This would should be run during install

4. Option for not by default overwriting
Install directory should be customizable /opt/datafed – default but needs to be flexible

PR #868, opened 2022-12-22
Should only check path for consistency with what is visible from Glob…

JoshuaSBrown:
…us collection root

As an example this is the content of path: file ftp://datafed-gcs-test/home/cades/collections/mapped/datafed-home/user/brownjs/1M.dat

Whereas in the "devel" branch this is the content of alloc.path + data_key: /datafed-home/user/brownjs/68722984

In "devel" the decision was made that the core server should not have insight into the full path of the POSIX filesystem, it only needs to know the directory structure of the collection root which it has control of. The Repo and Authz modules that exist on the same machine as the collection can check the actual POSIX path.

This fix also removes the use of the magick number 8 which appears to be dependent on how the repository directory hierarchy was set up and is brittle.

### DIFF
--- a/core/database/api/authz_router.js
+++ b/core/database/api/authz_router.js
@@ -66,7 +66,7 @@ router.get('/gridftp', function (req, res) {
         if ( !alloc )
             throw g_lib.ERR_PERM_DENIED;
 
-        if ( alloc.path + data_key != path ){
+        if ( ! path.endsWith(alloc.path + data_key) ){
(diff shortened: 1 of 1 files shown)


dvstans:
Recommend holding off on this - the changes have disabled certain aspects of gridftp access validation. Also may indicate deeper issues with tracking of data storage paths in DB allocation edges (used to form transfer request docs)

PR #873, opened 2023-01-06
Devel g5 fix

dvstans:
This PR changes the handling of the Globus 5 base path for repository endpoints. The base (and repo) paths are now set in the repo configuration in the DB and the repo server expects full paths to be sent to it. This fix has been tested on Globus 5; however, it will likely break Globus 4 installations. (No Globus 4 test server is available.) Note that Globus 4 is being deprecated soon.

### DIFF
--- a/common/proto/SDMS_Auth.proto
+++ b/common/proto/SDMS_Auth.proto
@@ -907,14 +907,13 @@ message RepoCreateRequest
     required string             id          = 1; // ID of repo
     required string             title       = 2; // Title
     optional string             desc        = 3; // Description
-    optional string             domain      = 5; // RESERVED
-    required string             path        = 6; // Path to storage directories
-    optional string             exp_path    = 7; // RESERVED
-    required string             address     = 8; // Repo server address
-    required string             endpoint    = 9; // Globus endpoint UUID or legacy name
-    required string             pub_key     = 10; // Public encryption key
-    required uint64             capacity    = 11; // Total data capacity
-    repeated string             admin       = 12; // Repo admin(s)
+    required string             base_path   = 4; // Path to base data collection (/path/to/data)
+    required string             repo_path   = 5; // Path to repo directory relative to base (/repo1/)
+    required string             address     = 6; // Repo server address
+    required string             endpoint    = 7; // Globus endpoint UUID or legacy name
+    required string             pub_key     = 8; // Public encryption key
+    required uint64             capacity    = 9; // Total data capacity
+    repeated string             admin       = 10; // Repo admin(s)
 }
 
 // Request to update an existing repository. Only system or repos admins may
@@ -924,16 +923,15 @@ message RepoCreateRequest
 message RepoUpdateRequest
 {
     required string             id          = 1; // ID of repo
-    optional string             title       = 2; // Title
+    required string             title       = 2; // Title
     optional string             desc        = 3; // Description
-    optional string             domain      = 5; // RESERVED
-    optional string             path        = 6; // Path to storage directories
-    optional string             exp_path    = 7; // RESERVED
-    optional string             address     = 8; // Repo server address
-    optional string             endpoint    = 9; // Globus endpoint UUID or legacy name
-    optional string             pub_key     = 10; // Public encryption key
-    optional uint64             capacity    = 11; // Total data capacity
-    repeated string             admin       = 12; // Repo admin(s)
+    required string             base_path   = 4; // Path to base data collection (/path/to/data)
+    required string             repo_path   = 5; // Path to repo directory relative to base (/repo1/)
+    required string             address     = 6; // Repo server address
+    required string             endpoint    = 7; // Globus endpoint UUID or legacy name
+    required string             pub_key     = 8; // Public encryption key
+    required uint64             capacity    = 9; // Total data capacity
+    repeated string             admin       = 10; // Repo admin(s)
 }
 
 // Request to delete a repository. Only system or repos admins may send this
--- a/core/database/api/repo_router.js
+++ b/core/database/api/repo_router.js
@@ -79,6 +79,57 @@ router.get('/view', function (req, res) {
 .summary('View repo server record')
 .description('View repo server record');
 
+function parseRepoDoc( repo ){
+
+    var obj = {
+        capacity: repo.capacity,
+        pub_key: repo.pub_key.trim(),
+        address: repo.address.trim(),
+        endpoint: repo.endpoint.trim(),
+        base_path: repo.base_path.trim(),
+        repo_path: repo.repo_path.trim()
+    };
+
+    g_lib.procInputParam( repo, "title", false, obj );
+    g_lib.procInputParam( repo, "summary", false, obj );
+    g_lib.procInputParam( repo, "domain", false, obj );
+
+    if ( obj.pub_key.length != 40 ){
+        throw [ g_lib.ERR_INVALID_PARAM, "Public key size is invalie (must be 40 characters)" ];
+    }
+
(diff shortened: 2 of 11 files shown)


core/database/api/repo_router.js
```diff
@@ -92,48 +143,26 @@ router.post('/create', function (req, res) {
...
                     throw g_lib.ERR_PERM_DENIED;
 
-                var obj = {
-                    capacity: req.body.capacity,
-                    pub_key: req.body.pub_key,
-                    address: req.body.address,
-                    endpoint: req.body.endpoint,
-                    path: req.body.path
-                };
+                var obj = parseRepoDoc( req.body );
```
JoshuaSBrown, line 146: I love changes like this, this makes things so much more testable.

core/database/api/repo_router.js
```diff
@@ -92,48 +143,26 @@ router.post('/create', function (req, res) {
...
-                        obj.path += "/";
-                }
 
                 var repo = g_db.repo.save( obj, { returnNew: true });
 
                 for ( var i in req.body.admins ) {
-                    if ( !g_db._exists( req.body.admins[i] ))
+                    if ( !g_db._exists( req.body.admins[i] )){
                         throw [g_lib.ERR_NOT_FOUND,"User, "+req.body.admins[i]+", not found"];
+                    }
```
JoshuaSBrown, line 155: This is an improvement, I have hit several errors when I have accidentally deleted white space in front of code that had no curly braces.

core/database/api/repo_router.js
```diff
@@ -92,48 +143,26 @@ router.post('/create', function (req, res) {
...
-                var obj = {
-                    capacity: req.body.capacity,
-                    pub_key: req.body.pub_key,
-                    address: req.body.address,
-                    endpoint: req.body.endpoint,
-                    path: req.body.path
-                };
+                var obj = parseRepoDoc( req.body );
 
                 g_lib.procInputParam( req.body, "id", false, obj );
```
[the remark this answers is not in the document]
dvstans: The ID field is only present in the create method not the update. Could have been checked in the parse function either way.

core/database/api/repo_router.js
dvstans: Yeah, I'm trying to fix these when I find them / think about it.

core/database/api/repo_router.js
```diff
@@ -146,18 +175,16 @@ router.post('/create', function (req, res) {
     id: joi.string().required(),
     title: joi.string().required(),
     desc: joi.string().optional(),
-    domain: joi.string().optional(),
-    capacity: joi.number().integer().min(0).required(),
+    capacity: joi.number().integer().min(1).required(),
```
JoshuaSBrown, line 178: What is the reason for this change? Are you just wanting to ensure that when someone creates a repo they actually assign something to it?

common/proto/SDMS_Auth.proto
```diff
@@ -907,14 +907,13 @@ message RepoCreateRequest
     required string             id          = 1; // ID of repo
     required string             title       = 2; // Title
     optional string             desc        = 3; // Description
-    optional string             domain      = 5; // RESERVED
-    required string             path        = 6; // Path to storage directories
-    optional string             exp_path    = 7; // RESERVED
```
JoshuaSBrown, line 912: I'm not following what "exp_path" is actually for from the Globus 4?

core/database/api/repo_router.js
dvstans: are you referring to the capacity? if so, a repo must have a non-zero capacity to make sense. The original min(0) was a mistake.

common/proto/SDMS_Auth.proto
dvstans: It was intended to be used in conjunction with the FUSE driver to tell end-users what the local path to the controlled data in the repo was (so they could open it directly). None of this will work going forward if we want to support non-posix storage options.

core/database/api/repo_router.js
JoshuaSBrown: Makes sense.

repository/server/RequestWorker.cpp
```diff
@@ -279,10 +281,11 @@ RequestWorker::procDataGetSizeRequest()
         }
         else
         {
-            data_sz->set_size( 0 );
```
JoshuaSBrown, line 282: Why did you remove this line?

dvstans: Odd, this is not showing the latest code changes, but it was replaced with an exception due to not finding the specified file path. The prior code would allow a transfer to succeed if there was a repo issue.

web/static/dlg_repo_edit.js
```diff
@@ -188,7 +184,7 @@ export function show( a_repo_id, a_cb ){
                     node.data.alloc.totalSz = data.totalSz;
                     updateAllocTitle( node );
 
-                    var msg =
+                    var msg =a_repo_id
```
JoshuaSBrown, line 187: So this looks like it is only sending the repo_id back is that the intention or is my javascript off?

dvstans: No, that's a weird typo. Not sure how that got there but needs to be fixed (there's a msg body below this not shown).

web/static/dlg_repo_edit.js
```diff
@@ -118,15 +114,15 @@ export function show( a_repo_id, a_cb ){
...
     if ( a_repo_id ){
-        $("#title",frame).on('input', function(){ repoInputChanged(1); });
-        $("#desc",frame).on('input', function(){ repoInputChanged(2); });
-        $("#domain",frame).on('input', function(){ repoInputChanged(4); });
-        $("#capacity",frame).on('input', function(){ repoInputChanged(8); });
-        $("#path",frame).on('input', function(){ repoInputChanged(0x10); });
-        $("#exp_path",frame).on('input', function(){ repoInputChanged(0x20); });
-        $("#pub_key",frame).on('input', function(){ repoInputChanged(0x40); });
-        $("#addr",frame).on('input', function(){ repoInputChanged(0x80); });
-        $("#ep_id",frame).on('input', function(){ repoInputChanged(0x100); });
```
JoshuaSBrown, line 129: This is definately cleaner and more readable getting rid of the magick numbers, though I'm not sure I fully grasp the implications of the change, I'm guessing you were using a bitmask to specify what parts needed to be updated and are now instead just doing a blanket update of the form?

repository/server/RequestWorker.cpp
JoshuaSBrown: Tyler and I noticed the other day that GitHub diffs were not showing the most recent changes perhaps this is the same problem.

web/static/dlg_repo_edit.js
dvstans: Yep, no need to support partial updates.

web/static/dlg_repo_edit.js
```diff
@@ -274,97 +270,60 @@ export function show( a_repo_id, a_cb ){
...
-                    if ( cap == null ){
-                        dialogs.dlgAlert("Data Entry Error","Invalid repo capacity value." );
-                        return;
-                    }
-                    if ( cap != repo.capacity )
-                        obj.capacity = cap;
+                    obj.id = a_repo_id;
+                } else {
+                    obj.id = $("#id",frame).val();
+                }
```
JoshuaSBrown, line 291: This is a lot clearer.

JoshuaSBrown (approved):
There is only one thing that still needs to be updated, there is a helper script that tells admins how to fill in the repo form in /scripts/globus it probably needs to be updated to reflect the new fields, but... its not really a blocking problem.

PR #875, opened 2023-02-02
Synack w004 19

dvstans:
This PR includes fixes for injection attack reported by Synack. There is one additional update in ClientWorker due to an update in the nlohmann json library. This change may not be backward compatible with older json libraries.


JoshuaSBrown:
I tested this and it works for me.

PR #883, opened 2023-09-18
Joshua s brown fix doc build

JoshuaSBrown:
Add script for installing documentation dependency and github action.


JoshuaSBrown:
@par-hermes format

PR #906, opened 2023-10-12
Add helper admin scripts

JoshuaSBrown:
# Description

This pr adds two scripts:

1. To create a backup of the arangodb database
2.  A second to update certificates

Both of the scripts are needed in the operation of DataFed. 

# Tasks

* [x] - Add labels
* [x] - Add reviewer
* [x] - Add assigned developer
* [x] - all tests pass
* [x] - add a description


JoshuaSBrown:
@par-hermes format

scripts/generate_datafed_backup_script.sh
nedvedba, line 1: In this script you are using `tar` to create a backup of the database when you likely could just use `arangodump` or `arangobackup` to do it in the intended fashion, which could also be restored with the builtin `arangorestore` as well. I anticipate that if you used one of those tool you might not have to stop and start the services to get a clean backup.

nedvedba (commented):
In the backup script you are using `tar` to create a backup of the database when you likely could just use `arangodump` or `arangobackup` to do it in the intended fashion, which could also be restored with the built-in `arangorestorev as well. I anticipate that if you used one of those tools you might not have to stop and start the services to get a clean backup.

dvstans:
> In the backup script you are using `tar` to create a backup of the database when you likely could just use `arangodump` or `arangobackup` to do it in the intended fashion, which could also be restored with the built-in `arangorestorev as well. I anticipate that if you used one of those tools you might not have to stop and start the services to get a clean backup.

It was my understanding that arangodump and arangobackup would not maintain consistency while the database was running under load. If this is the case, these may be better options.

nedvedba:
It seems like `arangodump` may be consistent while the database is running for newer versions but it is only a logical backup, `arangobackup` does support hot backups however you have to pay for Arango Enterprise so it may just be best to stick with forming an archive of the files for complete physical backups, unless it is decided to invest in enterprise edition in the future, in which case this could be revisited.

nedvedba (approved):

PR #919, opened 2023-11-03
Add retries as option when working with data transfers

JoshuaSBrown:
# Description

In the face of network faults, DataFed is canceling tasks. This needs to be controlled with an option. Instead, the default Globus behavior should be used, if a user specifically wants to cancel a transfer after a determined number of faults that is an option they can specify. This is particularly important when first testing a repository and attempting to detect network errors, but it is also useful to cancel tasks early in cases where there is high network traffic and an alternative transfer request to a different Globus endpoint might be preferable.

# Tasks

* [x] - Upgrade version
* [ ] - Changelog comment
* [ ] - Add tests
* [ ] - Feature description
* [ ] - Working tests in ci
* [ ] - Formatting passes
* [ ] - Documentation added
* [x] - Add labels
* [ ] - Assign work
* [x] - Assign expected time

# Work time

Expected time: 24 hours

### DIFF
--- a/common/CMakeLists.txt
+++ b/common/CMakeLists.txt
@@ -16,7 +16,8 @@ if( BUILD_COMMON )
     "source/operators/*.cpp"
     "source/servers/*.cpp"
     "source/support/zeromq/*.cpp"
-    "source/sockets/*.cpp")
+    "source/sockets/*.cpp"
+    "source/type_converters/*.cpp")
   add_library( common STATIC ${Sources})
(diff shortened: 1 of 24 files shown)


JoshuaSBrown:
Still need to address the faults logic in the GlobusAPI.cpp file and update the documentation. Also need to update the web server to support a retries parameter.

JoshuaSBrown:
@par-hermes format

AronPerez:
Not configurable, creates problems

PR #912, opened 2023-10-18
Joshua s brown feature add dev compose

JoshuaSBrown:
Adds initial compose file
Also adds docker file for building out python client
Removes end to end testing for now

PR #929, opened 2024-01-22
Blake nedved feature add dev compose

JoshuaSBrown:


compose.yaml
```diff
@@ -0,0 +1,22 @@
+version: '3'
+
+services:
+  web:
+    image: datafed-ws
+    depends_on:
+      - arangodb
+    ports:
+      - '80:80'
+      - '443:443'
```
JoshuaSBrown, line 10: The host ports will need to be configurable. Just keep that in mind. We can create an .env file with defaults in the same folder as the compose file to do this. Also, I recommend putting compose.yml in a compose folder instead of directly in the root of the project to keep things clean.

JoshuaSBrown:
This branch is being closed in favor or this:

https://github.com/ORNL/DataFed/pull/912

PR #957, opened 2024-05-28
Updated ArangoDB version to 3.12 for compose

nedvedba:
# Description

Updates the ArangoDB version used with Docker compose.


JoshuaSBrown:
You need to add labels, you also need to add a description. Also we should think of ways in which we can update this from the dependency versions file in the ./scripts folder. The version of arango is hard coded in multiple places. This needs to be reduced to a single place.

JoshuaSBrown:
We can actually consolidate the version numbers later in a separate PR, but for now at least add an item in the CHANGELOG

PR #958, opened 2024-06-05
initial fixes for worst files

nedvedba:
# Description

The point of this PR is to reduce the number of complaints coming from the static code analyzers thus making it clearer when actual errors or issues are reported.


JoshuaSBrown:
@nedvedba pull in the latest changes from devel and that should fix the CI

JoshuaSBrown:
@par-hermes format

core/database/foxx/api/support.js
```diff
@@ -1844,7 +1844,7 @@ module.exports = (function() {
                 }
 
                 // User ACL first
-                if (parent.acls && ((parent.acls & 1) != 0)) {
```
JoshuaSBrown, line 1847: @nedvedba do you understand what this is doing? I don't, it would be good to have an explanation of what is in parent.acls before approving this.

tests/end-to-end/test_api_alloc.py
```diff
@@ -51,16 +51,16 @@ def setUp(self):
...
                 break
             except BaseException:
                 pass
             count += 1
             # Try three times to authenticate
             assert count < 3
 
-        print("Attempt to login result")
-        print(result)
+        print("\nAttempted to login\n")
```
JoshuaSBrown, line 62: This is not the behavior we want, it would be beneficial to still print the results. I'm not sure why the analyzer was complaining about this.

JoshuaSBrown: Nevermind, the assertion should fail if there is a problem. In hind sight this seems fine.

JoshuaSBrown (approved):

PR #962, opened 2024-06-16
Add CI commit hash tag to docker images

JoshuaSBrown:
# Description

It looks like the image registry is not always being updated with the latest image from the build pipeline even though an entry can be found in the registry the image is missing. This can lead to build pipelines using the so called "latest" image for a branch which may not be the most up to date image. To fix this we can tag each image with the commit hash and use that in a given pipeline. The pipeline should at least fail if the image was not successfully uploaded. That way we don't actually use old images in a new test pipeline.


scripts/ci_harbor_artifact_count.sh
```diff
@@ -0,0 +1,97 @@
+#!/bin/bash
+
+# This script is created to determine if a harbor image exists in the repository
+# It will print the number of artifacts that exist for a particualar image.
+LOG_FILE="harbor_check.log"
+echo "CI Harbor Check Log File" > "$LOG_FILE"
+
```
AronPerez, line 7: Should we be using `set -u` flag for `nounset causes the shell to treat unset variables as an error and exit.`

.gitlab-ci.yml
```diff
@@ -20,6 +20,7 @@ stages:
 variables:
   HARBOR_USER: 'robot$$datafed+harbor_datafed_gitlab_ci_registry_2'
   REGISTRY: 'camden.ornl.gov'
+  DATAFED_HARBOR_REGISTRY: 'camden.ornl.gov'
```
AronPerez, line 23: Do we ever expect the `DATAFED_HARBOR_REGISTRY` to differ from `REGISTRY`? If so, we can use `${REGISTRY}`I believe

.gitlab-ci.yml
```diff
@@ -121,28 +122,52 @@ build-dependencies:
...
-    - docker push "${REGISTRY}/${IMAGE_TAG}-${BRANCH_LOWER}:latest"
+    - export DATAFED_HARBOR_REPOSITORY="${IMAGE_TAG}-${BRANCH_LOWER}"
+    - export DATAFED_HARBOR_USERNAME="${HARBOR_USER}"
+    - export DATAFED_HARBOR_PASSWORD="${HARBOR_DATAFED_GITLAB_CI_REGISTRY}"
+    - |
+      while [ "$(./scripts/ci_harbor_artifact_count.sh)" == "0" ]; do
+        echo "Artifact missing from harbor..."
+        docker push "${REGISTRY}/${DATAFED_HARBOR_REPOSITORY}:latest"
+        docker push "${REGISTRY}/${DATAFED_HARBOR_REPOSITORY}:$CI_COMMIT_SHA"
+        sleep 5  # Optional: Add a sleep to avoid busy waiting
```
AronPerez, line 142: This seems unnecessary and adding time unless I'm missing something

.gitlab-ci.yml
```diff
@@ -121,28 +122,52 @@ build-dependencies:
...
```
AronPerez, line 168: See comment above

.gitlab-ci.yml
```diff
@@ -192,15 +218,26 @@ build-ws:
...
+    - docker tag "${REGISTRY}/${IMAGE_TAG}-${BRANCH_LOWER}:latest" "${REGISTRY}/${IMAGE_TAG}-${BRANCH_LOWER}:$CI_COMMIT_SHA"
+    - export DATAFED_HARBOR_REPOSITORY="${IMAGE_TAG}-${BRANCH_LOWER}"
+    - export DATAFED_HARBOR_USERNAME="${HARBOR_USER}"
+    - export DATAFED_HARBOR_PASSWORD="${HARBOR_DATAFED_GITLAB_CI_REGISTRY}"
+    - |
+      while [ "$(./scripts/ci_harbor_artifact_count.sh)" == "0" ]; do
+        echo "Artifact missing from harbor..."
+        docker push "${REGISTRY}/${DATAFED_HARBOR_REPOSITORY}:latest"
+        docker push "${REGISTRY}/${DATAFED_HARBOR_REPOSITORY}:$CI_COMMIT_SHA"
+        sleep 5  # Optional: Add a sleep to avoid busy waiting
```
AronPerez, line 231: See comment above

.gitlab-ci.yml
```diff
@@ -219,15 +256,26 @@ build-core:
...
```
AronPerez, line 269: See comment above

.gitlab-ci.yml
AronPerez, line 142: Lots of duplicate code, can we follow the idea [here](https://github.com/ORNL/DataFed/pull/966#discussion_r1679638826) with a template and [extends](https://docs.gitlab.com/ee/ci/yaml/#extends)?

scripts/ci_harbor_artifact_count.sh
```diff
@@ -0,0 +1,97 @@
...
+then
+  echo "The DATAFED Harbor Username has not been defined." >> "$LOG_FILE"
+  exit 1
+fi
+
+if [ -z "$local_DATAFED_HARBOR_PASSWORD" ]
+then
+  echo "The DATAFED Harbor Password has not been defined." >> "$LOG_FILE"
+  exit 1
+fi
```
AronPerez, line 75: Would it be possible to make this a for loop?

```sh
local_DATAFED_HARBOR_REGISTRY=${DATAFED_HARBOR_REGISTRY:-""}
local_DATAFED_HARBOR_REPOSITORY=${DATAFED_HARBOR_REPOSITORY:-""}
local_DATAFED_HARBOR_USERNAME=${DATAFED_HARBOR_USERNAME:-""}
local_DATAFED_HARBOR_PASSWORD=${DATAFED_HARBOR_PASSWORD:-""}

# Check for required variables
for var in local_DATAFED_HARBOR_REGISTRY local_DATAFED_HARBOR_REPOSITORY local_DATAFED_HARBOR_USERNAME local_DATAFED_HARBOR_PASSWORD; do
  if [ -z "${!var}" ]; then
    echo "The ${var#local_} has not been defined." >> "$LOG_FILE"
    exit 1
  fi
done
```

We'd use [Shell Parameter Expansion](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html)

AronPerez (changes requested):
Nice identification and addressing here. Question, is this issue specific to Harbor?

A couple of action items:
1. Answering questions regarding variable assignment and flags set in shell scripts
2. Seeing if we can create a template for duplicate code and extending it to the necessary job.
3. Addressing sleep usage
4. Refactoring shell script via shell parameter expansion possibly

scripts/ci_harbor_artifact_count.sh
nedvedba: Bash has arrays that could make this even simpler, or at least look cleaner.

.gitlab-ci.yml
nedvedba: There are two registries but I can't think of a reason that both would have to be used at once.

scripts/ci_harbor_artifact_count.sh
JoshuaSBrown: @nedvedba not quite sure what that implementation might look like?

.gitlab-ci.yml
JoshuaSBrown: I was having issues with the registry processing the upload request before I verify that the image existed. I did want to overload the service with push requests while it was still processing, but I also wanted to be persistent until something was registered. I'm guessing that a better idea would be to not call sleep until the second round through the while loop.

JoshuaSBrown: Yes, that was addressed in #966

scripts/ci_harbor_artifact_count.sh
JoshuaSBrown: Nice catch.

.gitlab-ci.yml
JoshuaSBrown: This is more to be consistent and help organize the CI variables of which there are many. The prefixes help to organize the CI env variables according to their usage.

JoshuaSBrown: Oh I see the issue thanks for pointing this out.

JoshuaSBrown: One is a password, the other is the actual registry. I should probably change the env variable so it is clearer.

.gitlab/common.yml
```diff
@@ -20,31 +20,74 @@
...
     reports:
       dotenv: build.env
 
-
+# NOTE
+#
+# Cases have been observed where an image has been pushed to the image registry but 
+# but the image for some reason is not registered with the Harbor instance. The
+# retry loop is used to retry pushing to the registry to add some resliency to the
+# ci pipeline in the case of poor networking issues while attempting to upload
```
AronPerez, line 36: Very helpful to know 💯

AronPerez (approved):

.gitlab/common.yml
JoshuaSBrown: I'm glad you think so despite all the spelling and gramattical errors I'm now seeings :).

PR #964, opened 2024-07-02
Cheng web test

dingc2:
Adding web test


JoshuaSBrown:
Please add instructions for configuring and setting up Playwright.

dingc2:
Instructions to run playwright: 

Passwords:
In the auth.setup.js file, Passwords are currently accessed by storing a .env file in the web-UI directory. You can also place it in the environment, but make sure to un-require the dotenv node package in auth.setup.js 

Make sure the URL is the desired one in auth.setup.js, change the line to the correct datafed link:

In order to run the test, you have to install all the node packages:

npm ci 

Then install playwright as well as the browsers it uses 

npx playwright install

To run the test:

npx playwright test

JoshuaSBrown:
This is being closed in favor of this: https://github.com/ORNL/DataFed/pull/992

PR #965, opened 2024-07-09
Dylan xfer dlg

Dward944:
# Description

Aiming to alter the UI of the transfer dialog boxes, both upload and download. I am making this change to attempt to make the dialog boxes more intuitive and less clunky to use by rearranging its contents and removing unnecessary features. Currently, you go about using this dialog by inputting a path in the middle of the dialog, which then populates options in the drop down box above, which does not flow well. Additionally, the removal of the refresh and (re)activate buttons is due to them not functioning. 

## Current Dialog Boxes
image

## Planned Dialog Boxes
image 

## Implemented changes
image


web/static/dlg_start_xfer.js
```diff
@@ -81,206 +82,193 @@ export function show( a_mode, a_ids, a_cb ) {
...
-                }else{
-                    dialogs.dlgAlert("Globus Error", data );
+                    path_in.val(cur_ep.name + (cur_ep.default_directory ? cur_ep.default_directory : "/"));
+                    updateEndpointOptions(cur_ep);
+                } else {
+                    dialogs.dlgAlert("Globus Error", data);
                 }
             });
         }
     });
```
JoshuaSBrown, line 133: Put a comment above function describing what it is doing. If possible give an example of what the input fields should be and explain the expected behavior.

web/static/dlg_start_xfer.js
```diff
@@ -81,206 +82,193 @@ export function show( a_mode, a_ids, a_cb ) {
...
 
-        if ( ep.length == 0 ){
+        if (ep.length == 0) {
             ep_list = null;
-            matches.html( "<option disabled selected>No Matches</option>" );
-            matches.selectmenu("refresh");
-            matches.selectmenu("disable");
+            matches.html("<option disabled selected>No Matches</option>");
+            matches.prop("disabled", true);
             return;
```
JoshuaSBrown, line 212: Leave a comment block above this function explaining what it is doing.

JoshuaSBrown:
This has been moved into #986.

PR #966, opened 2024-07-11
Add ws ci job files

JoshuaSBrown:
# Description

Unfortunately, GitLab does not give us the ability to check if a docker image exists in a registry as part of the rules. This means downstream jobs dependent on the existence of an image will fail if the image was created. This PR seeks to remedy this problem by checking the image registry first and then picking a job that does not have a rules section and will always execute the image builds as opposed to one with rules that will only build an image if relevant files have been touched.


.gitlab-ci.yml
```diff
@@ -169,185 +170,342 @@ provision-client:
...
+      #  stage: build
+      #  tags:
+      #    - docker
+      #    - ci-datafed-client
+      #  script:
+      #    - BRANCH_LOWER=$(echo "$CI_COMMIT_REF_NAME" | tr '[:upper:]' '[:lower:]')
+      #    - docker system prune -f
+      #    - docker build -f python/docker/Dockerfile.python-client-base.ubuntu -t "${REGISTRY}/${IMAGE_TAG}-${BRANCH_LOWER}:latest" .
+      #    - docker login "${REGISTRY}" -u "${HARBOR_USER}" -p "${HARBOR_DATAFED_GITLAB_CI_REGISTRY}"
+      #    - docker push "${REGISTRY}/${IMAGE_TAG}-${BRANCH_LOWER}:latest"
```
nedvedba, line 508: Potentially unnecessary comment

gitlab_templates/force_build_repo_image.yml
```diff
@@ -0,0 +1,20 @@
+stages:
+  - build
+
+# The only difference between this file and the build_repo_image.yml is that
+# it does not have a rules section
```
nedvedba, line 5: You only included this comment in some of the files, I don't know if that was intentional or not.

nedvedba (commented):
Overall simple repetitive changes for the most part, the only thing I would add it maybe an explanation of what the CI pipeline would look like after these changes, probably with a picture to explain it better. I would also ask if all of the tasks in the CI file cannot be converted to separate configs, like how you did the container building in this PR, then they should potentially be swapped to scripts. to keep the CI file shorter.

gitlab_templates/build_ws_image.yml
```diff
@@ -0,0 +1,39 @@
...
+    - ./scripts/generate_datafed.sh
+    - docker login "${REGISTRY}" -u "${HARBOR_USER}" -p "${HARBOR_DATAFED_GITLAB_CI_REGISTRY}"
+    - docker build --build-arg DEPENDENCIES="${REGISTRY}/datafed/dependencies-${BRANCH_LOWER}:latest" --build-arg RUNTIME="${REGISTRY}/datafed/runtime-${BRANCH_LOWER}:latest" -f web/docker/Dockerfile -t "${REGISTRY}/${IMAGE_TAG}-${BRANCH_LOWER}:latest" .
+    - docker push "${REGISTRY}/${IMAGE_TAG}-${BRANCH_LOWER}"
+
+# Prevents downstream pipeline from being empty and triggering a failure, the
+# build-ws job above may or may not run because of the rules section, if there
+# are no jobs in the pipeline GitLab has the unfortunate behavior of reporting
+# the trigger as a failure. To prevent the pipeline from being empty the
+# placeholder job is created.
```
AronPerez, line 32: Looks like this is the problem you're describing 😞 : https://gitlab.com/gitlab-org/gitlab/-/issues/419293
I see a proposal [here](https://gitlab.com/groups/gitlab-org/-/epics/11447#proposal). I'd like you to see if any action was taken on this so we can avoid the placeholder and keep this nice work clean

AronPerez (changes requested):
Clean changes here 🧼 

A couple of items we want to address:
1. Making a template that [extends](https://docs.gitlab.com/ee/ci/yaml/#extends) to jobs (`check-x-image`, `run-x-build`). This lets us remove repetitive actions and only need to pass in variables.
2. See if skipped trigger pipeline failing issue was resolved or addressed
3. Verify if some job bits that were commented out are needed or not
4. Remove commented out items

gitlab_templates/build_ws_image.yml
JoshuaSBrown: Good catch.

JoshuaSBrown: Looks like strategy: mirror is not yet supported in our version of GitLab. Maybe we can refactor at some point in the future.

They don't even have it in their official documentation yet. 

https://docs.gitlab.com/ee/ci/pipelines/downstream_pipelines.html

Screenshot (16)

.gitlab-ci.yml
```diff
@@ -169,185 +170,342 @@ provision-client:
...
+  stage: build
+  trigger:
+    include:
+      - artifact: foxx_image.yml
+        job: check-foxx-image
+    strategy: depend
+  variables:
+    REGISTRY: "${REGISTRY}"
+    HARBOR_USER: "${HARBOR_USER}"
+    HARBOR_DATAFED_GITLAB_CI_REGISTRY: "${HARBOR_DATAFED_GITLAB_CI_REGISTRY}"
```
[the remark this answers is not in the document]
JoshuaSBrown: Yes, this is a good point, nice catch!

.gitlab-ci.yml
```diff
@@ -169,185 +170,342 @@ provision-client:
...
+        #    - docker
+        #    - ci-datafed-client
+        #  script:
+        #    - BRANCH_LOWER=$(echo "$CI_COMMIT_REF_NAME" | tr '[:upper:]' '[:lower:]')
+        #    - docker system prune -f
+        #    - docker build -f python/docker/Dockerfile.python-client-base.ubuntu -t "${REGISTRY}/${IMAGE_TAG}-${BRANCH_LOWER}:latest" .
+        #    - docker login "${REGISTRY}" -u "${HARBOR_USER}" -p "${HARBOR_DATAFED_GITLAB_CI_REGISTRY}"
+        #    - docker push "${REGISTRY}/${IMAGE_TAG}-${BRANCH_LOWER}:latest"
+        #
+        #
```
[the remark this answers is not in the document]
JoshuaSBrown: Actually, we should probably add it back into the pipeline until we are able to refactor and support a different python client build process. The content could be useful if it works even if it isn't ideal.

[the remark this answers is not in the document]
JoshuaSBrown: Nevermind, I don't want to mess with it.

.gitlab-ci.yml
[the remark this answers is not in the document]
JoshuaSBrown: Done, but I used includes not trigger.

JoshuaSBrown:
@par-hermes format

nedvedba (approved):
Looks good to me, this setup is much cleaner, other than optimizing the rules for CI tasks but I think that should be done in a separate PR once the CI is stable to speed it up.

PR #968, opened 2024-07-23
Add base path environment variable

nedvedba:
This PR serves to introduce the DATAFED_GCS_COLLECTION_BASE_PATH environment variable, due to some deployment issues discovered when I attempted to deploy the development version on my VM, this will allow a configuration option to be set that will specify to the GCS container where the base of the Globus collection is. I have tested this by wiping my VM and redeploying from scratch and it worked without any extra configuration. The reason that we cannot use the DATAFED_GCS_COLLECTION_ROOT_PATH environment variable is that the root path defines the root of the filesystem that the collection can see and base path is the POSIX path that defines the actual start of the filesystem in the container that the collection begins.


compose/all/compose.yml
```diff
@@ -130,6 +130,7 @@ services:
       DATAFED_REPO_ID_AND_DIR: "${DATAFED_REPO_ID_AND_DIR}"
       DATAFED_GLOBUS_SUBSCRIPTION: "${DATAFED_GLOBUS_SUBSCRIPTION}"
       DATAFED_GLOBUS_CONTROL_PORT: "${DATAFED_GLOBUS_CONTROL_PORT}"
+      DATAFED_GCS_COLLECTION_BASE_PATH: "/"
```
JoshuaSBrown, line 133: This needs to also be added to compose/repo/compose.yml, also you might want to import it from .env file instead of hardcoding it here. That will mean you need to make changes in the generate_env scripts.

JoshuaSBrown:
Can you add in the description the difference between the ROOT path and the base path in the description.

JoshuaSBrown:
https://jira.ccs.ornl.gov/browse/DATAFED-167

JoshuaSBrown:
Make sure you change how the volumes are mounted as well, this should reflect the ROOT_PATH not the BASE_PATH.

What really should be mounted is ${DATAFED_GCS_COLLECTION_ROOT_PATH}/${DATAFED_REPO_ID_AND_DIR}

JoshuaSBrown:
Example flows:

ROOT_PATH = /mnt/datafed/datafed-home
BASE_PATH = /mnt/datafed

What gets show in the Globus web app?
/datafed-home of collection will show content of datafed-home

What gets stored in DataFed in the path when registering
/datafed-home

What gets stored in authz config globus-base-path
/mnt/datafed
Because Globus will pass a path /mnt/datafed/datafed-home/user/bob/file we have to remove the base path before authorizing with DataFed so we should send /datafed-home/user/bob/file to Core service

What about repo config with globus var
Metadata service creates allocation on machine, sends /datafed-home/user/bob/file to repo service -> repo service must attach /mnt/datafed to path.

JoshuaSBrown:
@par-hermes format

JoshuaSBrown:
* I have successfully tested this with compose/all.
  - BASE path /mnt
  - ROOT path /mnt/datafed

JoshuaSBrown:
I have also tested using compose/metadata with compose/repo with BASE and ROOT set to /.

PR #973, opened 2024-08-01
Add error log output if end-to-end test fails

JoshuaSBrown:
If the CI end to end test fails we want to see the log output of all of the containers this pr addresses this problem.


JoshuaSBrown:
@par-hermes format

PR #978, opened 2024-08-07
Dylan main subscribe feature removed

Dward944:
I removed the subscribe function because it was both unimplemented and unused. I removed both the function itself and where it would appear in the main browse tab.

Before and after image of the cursor drop down action items.
image

The lines of code removed were 1282-1298, 1475, 1968, 2102, 2194.


JoshuaSBrown:
Closing this because content is moved into #988

PR #981, opened 2024-08-08
Joshua s brown schema html escape fix

JoshuaSBrown:
Adds the fix that Cheng found.


web/static/dlg_schema_list.js
```diff
@@ -51,7 +51,7 @@ function loadSchemas(){
                 for ( var i in data.schema ){
                     sch = data.schema[i];
                     //src.push({ title: sch.id + (sch.ver?"-"+sch.ver:"") + (sch.cnt?" (" + sch.cnt + ")":"") + (sch.ownNm?" " + sch.ownNm:"") + (sch.ownId?" (" + sch.ownId +")":""), key: sch.id + ":" + sch.ver });
-                    src.push({ title: sch.id + ":" + sch.ver + (sch.cnt?" (" + sch.cnt + ")":"") + (sch.ref?" (R)":""), own_nm: sch.ownNm, own_id: sch.ownId.substr(2), id: sch.id, ver: sch.ver, cnt: sch.cnt, ref: sch.ref, key: sch.id + ":" + sch.ver });
+                    src.push({ title: sch.id + ":" + sch.ver + (sch.cnt?" (" + sch.cnt + ")":"") + (sch.ref?" (R)":""), own_nm: util.escapeHTML(sch.ownNm), own_id: sch.ownId.substr(2), id: sch.id, ver: sch.ver, cnt: sch.cnt, ref: sch.ref, key: sch.id + ":" + sch.ver });
```
nedvedba, line 54: This looks good as long as this other fields cannot be targeted by the same type of attack.

JoshuaSBrown: The other pieces are generated internally. The name is the only piece that comes from an external source.

PR #982, opened 2024-08-08
Dylan {server_default} removed

Dward944:
Removed the {server_default} piece of the path_in.val variable when choosing an endpoint in the xfer dialog. A simple .replace function was placed after it populates to remove the unnecessary text.

image

 
image


JoshuaSBrown:
Moved to [#984]

PR #997, opened 2024-08-29
Fix docker push tag

JoshuaSBrown:
# Description

In installing a DataFed repo at Drexel, it was discovered that there were problems with the image containers. It turns out that the release CI pipelines were missing the :latest tag when pushing to the registry. This was causing the end-to-end tests to give false positives in the pipelines. It is fixed in the devel branch because the CI pipeline files were refactored. Merging this should cause the CI pipelines to reproduce the errors that Drexel and LeHigh were observing.

## Summary by Sourcery

Fix the CI pipeline to include the ':latest' tag when pushing Docker images to the registry, addressing issues with image versioning and ensuring accurate end-to-end test results.

CI:
- Add the ':latest' tag to Docker image pushes in the CI pipeline to ensure the latest version is correctly tagged and pushed to the registry.


JoshuaSBrown:
This was created off of a stale branch closing and opening a spearte branch with up to date commits.

PR #1014, opened 2024-10-28
ArangoDB script error behavior

JoshuaSBrown:
surround systemctl call with set +e and set -e to prevent incorrect erroring behavior

## Summary by Sourcery

Bug Fixes:
- Prevent script from exiting on non-zero exit code when checking the status of arangodb3.service.


JoshuaSBrown:
> Hey @JoshuaSBrown - I've reviewed your changes - here's some feedback:
> 
> **Overall Comments**:
> 
> * Consider using `systemctl is-active --quiet arangodb3.service` instead of temporarily disabling error checking. This is the preferred approach for scripting with systemctl.
> 

Setting --quiet does not do what we want because it doesn't output the "active" string that I am capturing and making descions on.

PR #1019, opened 2024-10-29
[CI] Add script to clean globus nodes and run

### LINKED ISSUE #1011: CI - Handle limits on number of nodes registered with Globus endpoint
# Description

Running globus connect server node setup several times in a row causes problems because there is a limit to the number of nodes that can be attached to a single Globus endpoint. 

It leads to the following error.

Error adding node key: Too many keys associated with endpoint.

# Acceptance

The CI runs successfully, before running the DataFed GCS instance the python certificate cleaning script is run. The DataFed GCS instance then runs without reporting the following problem.

Error adding node key: Too many keys associated with endpoint.

JoshuaSBrown:
Closes #1011


JoshuaSBrown:
@par-hermes format

PR #1032, opened 2024-11-04
Add a bit of debug context

theodore-s-beers:
I think it could be helpful to know at this point that what is being printed is the exit code of the preceding curl command. It should be 0, of course, but we saw another value (60) while troubleshooting an issue with an endpoint. It took me a while to figure out what it was referring to.

## Summary by Sourcery

Enhancements:
- Add context to debug output by specifying that the printed value is the curl command's exit code.


theodore-s-beers:
@JoshuaSBrown, should I add an entry to the changelog?

Issue #1053, opened 2024-11-07
[Foxx, CI] - Ensure CMake flag for running foxx tests is turned on


JoshuaSBrown:
# Description

To enable foxx tests the CMakeLists setting needs to be set.

# Acceptance

[ ] - Once the Foxx setup job runs the the cmake --test target will trigger the behavior to run the Foxx unit tests.

JoshuaSBrown:
#1031

JoshuaSBrown:
```bash
docker ps
CONTAINER ID   IMAGE                                                                       COMMAND                  CREATED         STATUS         PORTS     NAMES
2af867beeceb   camden.ornl.gov/datafed/foxx-1053-enable-foxx-tests-in-ci-pipeline:latest   "/bin/bash -c /dataf…"   7 seconds ago   Up 7 seconds             foxx-1053-enable-foxx-tests-in-ci-pipeline-2ae41dbb-pKB2UEDSd5
cloud@ci-datafed-globus2:~$ docker exec -it 2af867beeceb /bin/bash
root@2af867beeceb:/datafed/source# env | grep ENABLE
ENABLE_FOXX_TESTS=TRUE
root@2af867beeceb:/datafed/source# cat build/CMakeCache.txt | grep ENABLE
ENABLE_END_TO_END_API_TESTS:BOOL=OFF
ENABLE_END_TO_END_WEB_TESTS:BOOL=OFF
ENABLE_FOXX_TESTS:BOOL=True
ENABLE_UNIT_TESTS:BOOL=ON
root@2af867beeceb:/datafed/source#
```

PR #1054, opened 2024-11-07
1053 enable foxx tests in ci pipeline

JoshuaSBrown:
#1053

### DIFF
--- a/.gitlab/end_to_end.yml
+++ b/.gitlab/end_to_end.yml
@@ -45,6 +45,7 @@ end-to-end-foxx-setup:
     - echo "docker run -d \\" >> "${RUN_FILE}"
     - echo "--name \"foxx-${BRANCH_LOWER}-${CI_COMMIT_SHORT_SHA}-${random_string}\" \\" >> "${RUN_FILE}"
     - echo "-e DATAFED_ZEROMQ_SYSTEM_SECRET=\"$CI_DATAFED_ZEROMQ_SYSTEM_SECRET\" \\" >> "${RUN_FILE}"
+    - echo "-e ENABLE_FOXX_TESTS=\"TRUE\" \\" >> "${RUN_FILE}"
     - echo "-e DATAFED_DOMAIN=\"$CI_DATAFED_DOMAIN\" \\" >> "${RUN_FILE}"
     - echo "-e DATAFED_DATABASE_PASSWORD=\"$CI_DATAFED_DATABASE_PASSWORD\" \\" >> "${RUN_FILE}"
     - echo "-e DATAFED_DATABASE_IP_ADDRESS_PORT=\"$CI_DATAFED_DATABASE_IP_ADDRESS_PORT\" \\" >> "${RUN_FILE}"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,6 +35,8 @@ setting is turned on DataFed will build it's libraries as shared and try to
 link to shared libraries." OFF)
 OPTION(ENABLE_END_TO_END_API_TESTS "Enable end-to-end API testing" FALSE) 
 OPTION(ENABLE_END_TO_END_WEB_TESTS "Enable end-to-end web testing with Playwright" FALSE)
+OPTION(ENABLE_FOXX_TESTS "Enable Foxx testing, off by default because it
+  will overwrite the test database." FALSE)
 
 set(INSTALL_REPO_SERVER ${BUILD_REPO_SERVER})
 set(INSTALL_AUTHZ ${BUILD_AUTHZ})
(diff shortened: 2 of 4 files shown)


.gitlab/end_to_end.yml
```diff
@@ -45,6 +45,7 @@ end-to-end-foxx-setup:
     - echo "docker run -d \\" >> "${RUN_FILE}"
     - echo "--name \"foxx-${BRANCH_LOWER}-${CI_COMMIT_SHORT_SHA}-${random_string}\" \\" >> "${RUN_FILE}"
     - echo "-e DATAFED_ZEROMQ_SYSTEM_SECRET=\"$CI_DATAFED_ZEROMQ_SYSTEM_SECRET\" \\" >> "${RUN_FILE}"
+    - echo "-e ENABLE_FOXX_TESTS=\"TRUE\" \\" >> "${RUN_FILE}"
```
AronPerez, line 48: So I understand the context, Foxx tests involve `ArangoDB` (looking at [test_foxx.sh](https://github.com/ORNL/DataFed/blob/9e9beda91bc29409673ca0bfe3f3e07cc4ebc941/core/database/tests/test_foxx.sh#L107))?

docker/entrypoint_foxx.sh
```diff
@@ -33,17 +33,32 @@ then
...
+      -DBUILD_REPO_SERVER=False   \
+      -DBUILD_COMMON=False        \
+      -DBUILD_AUTHZ=False         \
+      -DBUILD_CORE_SERVER=False   \
+      -DBUILD_WEB_SERVER=False    \
+      -DBUILD_DOCS=False          \
+      -DBUILD_PYTHON_CLIENT=False \
+      -DBUILD_FOXX=True           \
+      -DINSTALL_FOXX=True
+  fi
```
AronPerez, line 61: if `-DENABLE_FOXX_TESTS=True` is the only difference between `ENABLE_FOXX_TESTS` being true and false, can you do something like this with an inplace variable update?


```shell
cmake_options="-S. -B build \
  -DBUILD_REPO_SERVER=False   \
  -DBUILD_COMMON=False        \
  -DBUILD_AUTHZ=False         \
  -DBUILD_CORE_SERVER=False   \
  -DBUILD_WEB_SERVER=False    \
  -DBUILD_DOCS=False          \
  -DBUILD_PYTHON_CLIENT=False \
  -DBUILD_FOXX=True           \
  -DINSTALL_FOXX=True"

if [ "$ENABLE_FOXX_TESTS" == "TRUE" ]; then
  cmake_options="$cmake_options -DENABLE_FOXX_TESTS=True"
fi

"${DATAFED_DEPENDENCIES_INSTALL_PATH}/bin/cmake" $cmake_options
```

AronPerez (commented):
Looking good, some context questions around:
1. Foxx tests involve ArangoDB (looking at [test_foxx.sh](https://github.com/ORNL/DataFed/blob/9e9beda91bc29409673ca0bfe3f3e07cc4ebc941/core/database/tests/test_foxx.sh#L107))?
2. Overwriting the test database..?

Question around simplifying code:
1.  Setting flags on `"${DATAFED_DEPENDENCIES_INSTALL_PATH}/bin/cmake" -S. -B build`

CMakeLists.txt
```diff
@@ -35,6 +35,8 @@ setting is turned on DataFed will build it's libraries as shared and try to
 link to shared libraries." OFF)
 OPTION(ENABLE_END_TO_END_API_TESTS "Enable end-to-end API testing" FALSE) 
 OPTION(ENABLE_END_TO_END_WEB_TESTS "Enable end-to-end web testing with Playwright" FALSE)
+OPTION(ENABLE_FOXX_TESTS "Enable Foxx testing, off by default because it
+  will overwrite the test database." FALSE)
```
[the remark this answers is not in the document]
JoshuaSBrown: The commands you are looking at are only partly a problem. Meaning inserting a daily message won't really cause any problems. The second line is overwriting the config document. It's possible you could execute this in prod without causing problems, but it would be risky. 

The comment is perhaps a bit unclear. My concern was with creating and destroying repos, allocations and users etc on a production database as a result of running the tests.

.gitlab/end_to_end.yml
JoshuaSBrown: That's correct, Arangodb has a bit of weird feature that allows you to write JavaScript and run it like database stored procedure, it supports transactions. That's what foxx is.

docker/entrypoint_foxx.sh
JoshuaSBrown: How would you feel about me putting this in a follow up PR?

JoshuaSBrown: https://github.com/ORNL/DataFed/issues/1063

nedvedba (approved):
Approving on the condition that #1063 is successfully merged in.

Issue #1065, opened 2024-11-08
[CI] - Bug in the ci_purge script


JoshuaSBrown:
# Description

The CI job that is meant to purge is displaying this:

```
$ ./scripts/ci_purge_images.sh
docker_size_stats
Image size is 35.7GB
Cleaning up project directory and file based variables
00:00
Job succeeded
```

In addition, it won't run if the disk is found to be full.

```
Resolving secrets
Preparing the "shell" executor
00:00
Using Shell (bash) executor...
Preparing environment
00:00
Running on ci-datafed-core...
Getting source from Git repository
00:00
bash: line 5[4](https://code.ornl.gov/dlsw/datafed/datafed/-/jobs/2812077#L4): printf: write error: No space left on device
Cleaning up project directory and file based variables
00:00
ERROR: Job failed: exit status 1
```

Note it is only displaying the first print statement and then exiting. Running the script directly in the same VM lead to the following.

UPDATE - this is because in the CI the env var is set to allow up to 35 GB whereas the default in the purge script is set to 15GB so when run directly it is more aggresive.

```
./ci_purge_images.sh
docker_size_stats
Image size is 35.98GB
Untagged: camden.ornl.gov/datafed/repo-1053-enable-foxx-tests-in-ci-pipeline:1aa1d3eb78fca57b73364da2ce5a36c034677fac
Untagged: camden.ornl.gov/datafed/repo-1053-enable-foxx-tests-in-ci-pipeline:2ae41dbb23ce42c1ac9b6e6c72e8099d47e2e8e9
Untagged: camden.ornl.gov/datafed/repo-1053-enable-foxx-tests-in-ci-pipeline:latest
Untagged: camden.ornl.gov/datafed/repo-1053-enable-foxx-tests-in-ci-pipeline@sha256:89af00fa7d069670f2447231348781f49bab620487b4931dfcfa5a78fe3bc550
Untagged: camden.ornl.gov/datafed/repo-1060-output-from-running-arango-setup-container-to-stdout:038ac0552490b0450371c5993cfc7b8ea95d5abb
Untagged: camden.ornl.gov/datafed/repo-1060-output-from-running-arango-setup-container-to-stdout:latest
Untagged: camden.ornl.gov/datafed/repo-1060-output-from-running-arango-setup-container-to-stdout@sha256:89af00fa7d069670f2447231348781f49bab620487b4931dfcfa5a78fe3bc550
[...]

PR #1068, opened 2024-11-11
1065 fix ci cleanup script

JoshuaSBrown:
It was discovered that the CI jobs were not correctly clearing out old docker images. This was found to be a result of incorrectly assigned variables. This PR attempts to fix the purge script.

#1065

## Summary by Sourcery

Fix the CI cleanup script to correctly clear out old Docker images by addressing variable assignment issues and improving script robustness.

Bug Fixes:
- Fix the CI cleanup script to correctly clear out old Docker images by addressing variable assignment issues.

CI:
- Update the CI purge script to ensure proper handling of environment variables and improve script robustness with 'set -euf'.


nedvedba (approved):
Looks good to me, I would potentially think about implementing the changes sourcery suggested about not using tail as it may not guarantee getting the oldest chronological image, but that may only need to be addressed if we run into an issue.

PR #1125, opened 2024-11-21
Remove overwrite of PATH

JoshuaSBrown:
## Summary by Sourcery

Deployment:
- Remove the overwrite of the PATH environment variable in the Dockerfile for the Foxx service.

### DIFF
--- a/docker/Dockerfile.foxx
+++ b/docker/Dockerfile.foxx
@@ -22,10 +22,15 @@ ARG BUILD_DIR
 ARG DATAFED_INSTALL_PATH
 ARG DATAFED_DEPENDENCIES_INSTALL_PATH
 
+# WARNING
+#
+# Do not overwrite PATH here, overwriting PATH could lead to inconsistencies in
+# your build. Docker will automatically pull from the host env if declared here.
+# If PATH is different on the host you could create a broken container.
+
 ENV                         BUILD_DIR="${BUILD_DIR}"
 ENV                       DATAFED_DIR="${DATAFED_DIR}"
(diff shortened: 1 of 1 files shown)


docker/Dockerfile.foxx
```diff
@@ -22,10 +22,15 @@ ARG BUILD_DIR
...
+# WARNING
+#
+# Do not overwrite PATH here, overwriting PATH could lead to inconsistencies in
+# your build. Docker will automatically pull from the host env if declared here.
+# If PATH is different on the host you could create a broken container.
+
 ENV                         BUILD_DIR="${BUILD_DIR}"
 ENV                       DATAFED_DIR="${DATAFED_DIR}"
 ENV DATAFED_DEPENDENCIES_INSTALL_PATH="${DATAFED_DEPENDENCIES_INSTALL_PATH}"
-ENV                              PATH=${DATAFED_DEPENDENCIES_INSTALL_PATH}/bin:${PATH}
```
t-ramz, line 28: Why was the path being overwritten in the first place?

t-ramz (approved):

AronPerez (approved):

docker/Dockerfile.foxx
JoshuaSBrown: So that we didn't have to explicitly use /opt/datafed/dependicies/bin in front of the executables we were installing. It was an attempt to make the code more readable.

The reason the CI didn't catch it I think is because the container env was the same as the host, both are run in ubuntu.

PR #1160, opened 2024-12-04
1106 updated proto messages for setting token

nedvedba:
## Summary by Sourcery

Enhancements:
- Update proto messages in SDMS_Auth.proto to support setting tokens.


common/proto/common/SDMS_Auth.proto
```diff
@@ -116,6 +121,8 @@ message UserSetAccessTokenRequest
     required string             access      = 1; // Access token
     required uint32             expires_in  = 2; // Access expiration time in seconds
     required string             refresh     = 3; // Refresh token (never expires)
+    optional AccessTokenType    type        = 4;
```
t-ramz, line 124: Would this mean that optional params can be added as part of the object in the request here?

https://github.com/ORNL/DataFed/blob/devel/web/datafed-ws.js#L1529

common/proto/common/SDMS_Auth.proto
```diff
@@ -21,6 +21,11 @@ enum Protocol
     ID = 2;
 }
 
+enum AccessTokenType {
+    GENERIC = 1;
+    GLOBUS = 2;
+}
```
JoshuaSBrown, line 27: @t-ramz  @AronPerez @nedvedba I thought we talked about expanding on this in Globus with different types? i.e. GLOBUS_AUTH vs something else?

common/proto/common/SDMS_Auth.proto
JoshuaSBrown: Yes, exactly!

JoshuaSBrown (approved):
This looks good to me.

AronPerez (approved):

t-ramz:
It looks like this is causing some kind of error when logging in (which makes me think it is directly related).

JoshuaSBrown:
I'm so confused, the CI pipelines for this are missing the wait job, which enforces serialization of the pipelines. It's probably why the CI keep failing when it is rerun, it is interfereing with other pipelines.

JoshuaSBrown:
So this was failing because it was branched off of an old version of devel. Not sure how that happened.

JoshuaSBrown:
Looks like this is a result of "--dump" flag that is not valid. Not sure why this wasn't a problem before.

JoshuaSBrown:
```
Install the project...
-- Install configuration: ""
PROJECT ROOT /datafed/source
curl: option --dump: is unknown
curl: try 'curl --help' or 'curl --manual' for more information
Output: 
curl command failed [MASKED]/_api/database exiting
Running tests...
```

JoshuaSBrown:
@t-ramz can you try and take on #1172 and #1173, that should fix this issue.

common/proto/common/SDMS.proto
```diff
@@ -86,6 +86,14 @@ enum ProjectRole
     PROJ_ADMIN      = 3;
 }
 
+// Used to specify type of access token contained within message
+enum AccessTokenType {
```
t-ramz, line 90: This enum needed to be moved to SDMS.proto because defining enums in SDMS_Auth and SDMS_Anon cause issues when datafed-ws.js is importing messages with its current logic

JoshuaSBrown (approved):

PR #1176, opened 2024-12-11
Remove problematic curl dump flag

JoshuaSBrown:
# PR Description

# Tasks

## Summary by Sourcery

Bug Fixes:
- Remove the problematic '--dump' flag from the curl command in the test setup script.

### DIFF
--- a/scripts/dependency_install_functions.sh
+++ b/scripts/dependency_install_functions.sh
@@ -91,7 +91,7 @@ install_cmake() {
     # Mark cmake as installed
     touch "${DATAFED_DEPENDENCIES_INSTALL_PATH}/.cmake_installed-${DATAFED_CMAKE_VERSION}"
   fi
-  export PATH="${DATAFED_DEPENDENCIES_INSTALL_PATH}/bin:${PATH}"
+  export PATH="${DATAFED_DEPENDENCIES_INSTALL_PATH}/bin:$PATH"
 }
 
 install_protobuf() {
--- a/scripts/dependency_install_functions.sh
+++ b/scripts/dependency_install_functions.sh
@@ -49,6 +49,11 @@ else
   fi
 fi
 
+# WARNING: overwriting PATH can be very dangerous
+#   In Docker builds this must follow the pattern:
+#     PATH="<desired addition to path>:$PATH"
+#     Curly braces around PATH, like ${PATH} may pull from the host's PATH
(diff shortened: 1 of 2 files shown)


t-ramz:
While rebuilding after #1175 was fixed, I reached this error in the CI:
image


Which looks eerily familiar to #1125 

I have identified [line 94 in dependency_install_functions.sh](https://github.com/ORNL/DataFed/blob/devel/scripts/dependency_install_functions.sh#L94) as a possible culprit, maybe related to this SO answer:
https://stackoverflow.com/a/38742545

I am working to test that this is the case on a dev machine, and will push an update if it looks to be fixed.

JoshuaSBrown:
Nice work @t-ramz. 

This could also be due to missing docker layers. We are running scripts to check that there is enough hard drive space on the VMs. If docker system prune does not remove enough to guarantee space full the full build pipeline the oldest images are forcefully removed. Its possible that this could be part of the problem.

JoshuaSBrown:
Looks like the CI pipeline went through. Did you rerun the failed job? If you did it's possible it ran on a different VM where the build cache is fine.

scripts/dependency_install_functions.sh
```diff
@@ -91,7 +91,7 @@ install_cmake() {
     # Mark cmake as installed
     touch "${DATAFED_DEPENDENCIES_INSTALL_PATH}/.cmake_installed-${DATAFED_CMAKE_VERSION}"
   fi
-  export PATH="${DATAFED_DEPENDENCIES_INSTALL_PATH}/bin:${PATH}"
+  export PATH="${DATAFED_DEPENDENCIES_INSTALL_PATH}/bin:$PATH"
```
JoshuaSBrown, line 94: Why did you change this? I don't think this would be the source of the runc problem.

t-ramz: Last time we ran into an issue with the tail of the problem `stat /bin/bash: no such file or directory`, the issue was related to PATH being overwritten like in #1125 when building DataFed. I'm otherwise unsure how `bash` would ever be missing.

t-ramz:
I am going to make sure the branch is caught up with devel and work from there to see if there's something we can do.

t-ramz:
> Can we document this in a separate issue and address it there

Yessir 🫡 
#1177

AronPerez (approved):

PR #1186, opened 2024-12-17
Changes to allow for and auto generate debug log level

### LINKED ISSUE #1185: [CI, Build] Introduce the ability to set log level in a config file
## Description
When building DataFed, the application has a [hard-coded log level of LogLevel::INFO](https://github.com/ORNL/DataFed/blob/devel/core/server/main.cpp#L33).

This should be changed so that, given a config file, the config file can set this log level.

This will need a change here: https://github.com/ORNL/DataFed/blob/devel/core/server/main.cpp#L33 to allow for a new option.

- [ ] Update [config file generation](https://github.com/ORNL/DataFed/blob/devel/scripts/generate_core_config.sh#L195) to provide a default of DEBUG and specify log levels
- [ ] Update [config file documentation](https://github.com/ORNL/DataFed/blob/devel/docs/_sources/admin/general.rst.txt?plain=1#L183) to specify log levels

## Acceptance

- [ ] CI passes and logs at the DEBUG level

t-ramz:
# PR Description

Closes #1185 
Added capability to accept `log-level` as a configuration option.

# Tasks

# Artifacts
Logging with config file `log-level` set to 5 (TRACE):
image

Logging with config file `log-level` set to 50, invalid:
image

Logging with config file `log-level` set to 2 (WARNING):
image

## Summary by Sourcery

Add a configurable log level feature to the application, allowing users to specify the log level through configuration and command-line options. Update the documentation to reflect this new feature.

New Features:
- Introduce a configurable log level setting for the application, allowing users to set the log level from critical to trace.

Enhancements:
- Add a command-line option to set the log level in the server configuration.

Documentation:
- Update the admin documentation to include information about the new log level configuration option.

## Summary by Sourcery

Add a configurable log level feature to the application, allowing users to specify the log level through configuration and command-line options. Update the documentation to reflect this new feature.

New Features:
- Introduce a configurable log level setting for the application, allowing users to set the log level from critical to trace.

Enhancements:
- Add a command-line option to set the log level in the server configuration.

Documentation:
- Update the admin documentation to include information about the new log level configuration option.


core/server/main.cpp
```diff
@@ -15,6 +15,7 @@
 
 // Standard includes
 #include "Config.hpp"
+#include <climits>
```
t-ramz, line 18: This is included for access to `UINT_MAX`

core/server/main.cpp
```diff
@@ -147,6 +151,42 @@ int main(int a_argc, char **a_argv) {
...
+          case LogLevel::INFO:
+            str_log_level = "INFO";
+            break;
+          case LogLevel::DEBUG:
+            str_log_level = "DEBUG";
+            break;
+          case LogLevel::TRACE:
+            str_log_level = "TRACE";
+            break;
+          case LogLevel::LAST_SENTINEL:
```
t-ramz, line 181: Should I throw an exception in this case?

core/server/main.cpp
```diff
@@ -147,6 +151,42 @@ int main(int a_argc, char **a_argv) {
 
         return 0;
       }
+      if (cfg_log_level != UINT_MAX) {
+        if (cfg_log_level >=
+            static_cast<unsigned int>(LogLevel::LAST_SENTINEL)) {
+          DL_ERROR(log_context,
```
t-ramz, line 157: Should this throw an error instead of logging and defaulting?

t-ramz (commented):
Some questions for consideration

core/server/main.cpp
AronPerez: Yes

scripts/generate_core_config.sh
```diff
@@ -208,6 +208,14 @@ db-pass=$local_DATAFED_DATABASE_PASSWORD
...
 client-id=$local_DATAFED_GLOBUS_APP_ID
 client-secret=$local_DATAFED_GLOBUS_APP_SECRET
+# Below sets the log level of the application
+# 0 - Critical
+# 1 - Error
+# 2 - Warning
+# 3 - Info  (Default)
+# 4 - Debug
+# 5 - Trace
+log-level=4
```
AronPerez, line 218: Do we want the default to be 3 or 4?

```shell
3 - Info  (Default)
```

core/server/main.cpp
JoshuaSBrown: I think a default is fine, logs are not critical to the running of the service.

docs/_sources/admin/general.rst.txt
```diff
@@ -191,6 +191,7 @@ Example datafed-core.cfg file::
     cred-dir = /opt/datafed/keys
     client-id = <Globus App ID>
     client-secret = <Globus App Secret>
+    log-level = 4
```
JoshuaSBrown, line 194: Glad you added this! Good attention to detail.

scripts/generate_core_config.sh
JoshuaSBrown: Should be Info, you don't want people generating more logs than they need on a production system just because they didn't pay attention to the log settings.

JoshuaSBrown (approved):

core/server/main.cpp
```diff
@@ -147,6 +151,42 @@ int main(int a_argc, char **a_argv) {
...
+          case LogLevel::TRACE:
+            str_log_level = "TRACE";
+            break;
+          case LogLevel::LAST_SENTINEL:
+            str_log_level = "INVALID"; // should never be reached
+            break;
+          }
+          std::string log_message = "Setting log level to " + str_log_level;
+          DL_INFO(log_context, log_message);
+          global_logger.setLevel(cast_log_level);
```
JoshuaSBrown, line 187: Actually won't this be confusing if str_log_level is INVALID but we are defaulting to INFO?

JoshuaSBrown: Is the last case just there in case someone changes th code, will it be hit ever since you have an else checking that the SENTINEL value is not used?

scripts/generate_core_config.sh
t-ramz: How should I address this for the CI? The image is built using this script and it is run in the entrypoint, so I'm not sure how to overwrite it.

core/server/main.cpp
t-ramz: Yes! I changed it to throw an exception in case we hit an invalid state

core/server/main.cpp
t-ramz: Added 🫡

core/server/main.cpp
t-ramz: I added the last case so that every case is explicitly defined, and if the `LogLevel` enum is ever changed, a compiler error will be thrown to update the switch as well. I'm not sure this is my favorite approach, but it will give some extra safety over `default`

AronPerez (approved):

core/server/main.cpp
```diff
@@ -46,6 +47,10 @@ int main(int a_argc, char **a_argv) {
     string cfg_file;
     bool gen_keys = false;
 
+    // Note: we may want to dynamically choose type at compile time
```
t-ramz, line 50: I have an idea on how to implement this already if we would like to proceed sooner rather than later:
utilize `underlying_type` and `numeric_limits::max()` to retrieve the type and its respective max that we can use as a sentinel.

PR #1187, opened 2024-12-18
1184 bug add set e foxx install

JoshuaSBrown:
# PR Description

This PR was about getting the install_foxx to correctly throw an error when an error was detected. That error should cause the CI job to exit with an error code. This was correctly demonstrated in comm [3ec6d53](https://github.com/ORNL/DataFed/pull/1187/commits/3ec6d533e7b2cc9c41f9ae3c6190d78e878510c5). Then the fix was merged into this branch to show how it is working and the job is now passing, this was done by removing the curl --dump flag. 

# Tasks

* [x] - A description of the PR has been provided, and a diagram included if it is a new feature.
* [x] - Labels have been assigned to the pr
* [x] - A reviwer has been added
* [x] - A user has been assigned to work on the pr

Issue #1127, opened 2024-11-21
[Core, Web, Database] - Feature allow relay of tokens to database


JoshuaSBrown:
# Depends on 

https://github.com/ORNL/DataFed/issues/1109
https://github.com/orgs/ORNL/projects/7/views/1?pane=issue&itemId=88159403&issue=ORNL%7CDataFed%7C1106

# Description

When supporting mapped collections we need to support specifying the collection uuid when setting the token. That will require changes in the Client worker which translates protobuf to HTTP

```
void DatabaseAPI::userSetAccessToken(const std::string &a_acc_tok,
                                       uint32_t a_expires_in,
                                       const std::string &a_ref_tok,
                                       LogContext log_context) {
    string result;
    dbGetRaw("usr/token/set",
             {{"access", a_acc_tok},
              {"refresh", a_ref_tok},
              {"expires_in", to_string(a_expires_in)}},
             result);
    DL_TRACE(log_context, "token expires in: " << to_string(a_expires_in));
  }
```

# Acceptance

if a collection uuid is provided an http call to the database should include the collection uuid otherwise it shouldn't. This requires getting the user_router.js route that handles the tokens to accept collections uuids as an optional parameter.

JoshuaSBrown:
#1109 needs to be completed first.

Issue #1120, opened 2024-11-19
[Core, Database, Globus] Fetch and store new tokens to access mapped collections


t-ramz:
# Description
Depends on #1119 
The goal of this ticket is to support #1116 by collecting and storing new tokens that are required to access mapped collections

# Context
DataFed needs to provide access to mapped collections, which requires gathering and utilizing consent of a User. The Globus Auth API provides a method for generating new tokens based on a temporary access code returned when a user provides consent. These tokens should be stored and mapped to a user, with the context of the provided consent. This may be done as an additional column on the database table (or equivalent link/edge on a graph database).

t-ramz:
I was able to verify the token being placed on the queue through it being fetched from the queue, here is some limited output for it:
![Image](https://github.com/user-attachments/assets/8e572e36-5410-432d-b7ec-d83afd4f75f8)

PR #1194, opened 2024-12-20
Feature DLT 1109 foxx new token data

### LINKED ISSUE #1109: [Database, Foxx] Feature update the Foxx API, such that the token/set api call supports additional token info
# Description

If we want to have more than the default token used with DataFed, we need to be able to specify context such as the token type and any other information related to the token, so for instance with Globus we want the UUID the token can be used with.

```
router.get('/token/set', function(req, res) {
        try {
            g_db._executeTransaction({
                collections: {
                    read: ["u", "uuid", "accn"],
                    write: ["u"]
                },
                action: function() {
                    const client = g_lib.getUserFromClientID(req.queryParams.client);
                    var user_id;

                    if (req.queryParams.subject) {
                        user_id = req.queryParams.subject;
                        if (!g_db.u.exists(user_id))
                            throw [g_lib.ERR_INVALID_PARAM, "No such user '" + user_id + "'"];
                        g_lib.ensureAdminPermUser(client, user_id);
                    } else {
                        user_id = client._id;
                    }
                    console.log("updating tokens for", user_id, "acc:", req.queryParams.access, "exp:", req.queryParams.expires_in);
                    var obj = {
                        access: req.queryParams.access,
                        refresh: req.queryParams.refresh,
                        expiration: Math.floor(Date.now() / 1000) + req.queryParams.expires_in
                    };
                    g_db._update(user_id, obj, {
                        keepNull: false
                    });
                }
            });
        } catch (e) {
            g_lib.handleException(e, res);
        }
    })
    .queryParam('client', joi.string().required(), "Client ID")
    .queryParam('subject', joi.string().optional(), "UID of subject user")
    .queryParam('access', joi.string().required(), "User access token")
    .queryParam('refresh', joi.string().required(), "User refresh token")
    .queryParam('expires_in', joi.number().integer().required(), "Access token expiration timestamp")
    .summary('Set user tokens')
    .description('Set user tokens');
 ```
 
 This call in the foxx user_router needs to be changed to accept queryParams where the token type can be specified along with the collection UUID.
 
 # Acceptance
 
 - [x] A test is created that calls the user_router token/set with a GET request, this tests validates that without specifying token type or additional info the call works as expected
 - [x] A test is created that calls user_router token/set and a GET with TOKEN type supported specified as the Globus token type and with additional information about the globus collection uuid it is handled correctly and stored as an edge connected to the user.

### PR DESCRIPTION
**Pull request by t-ramz**
Closes #1109 

# PR Description
Add support for new token options to Arango Foxx API.
Add and support testing user router with database fixtures.

Code changes here allow for acceptance and usage of new token options in SDMS::UserSetAccessTokenRequest.
These changes will build new documents in the previously unused `globus_coll` collection if they are not present that will represent a Globus Collection. The changes will also relate the `globus_coll` collection to the `u` user collection via the `globus_token` edge where tokens will be stored for accessing said Globus collections.

# Artifacts
Resulting data from tests in `globus_coll`
image

Resulting data from tests in `globus_token`
image

Resulting user from test same as above
image

# Tasks

* [x] - CHANGELOG comment has been added
* [x] - If new feature a unit test has been added

## Summary by Sourcery

Add support for storing different token types in the Arango Foxx API, by adding new fields to the `/token/set` endpoint.  A new edge collection `globus_token` is introduced to store tokens associated with a specific Globus collection. Update the user fixture script to create a test user for Foxx tests.

New Features:
- Added support for new token types in the Arango Foxx API.

Tests:
- Added integration tests for the new token options in the Foxx user router.

### CODE REVIEW

File: core/database/foxx/api/user_router.js [hunk 1]
```diff
@@ -516,9 +519,47 @@ router.get('/token/set', function(req, res) {
...
+                                                                    //  which should be kicked out by arango
+                        if (!user_doc) {
+                            throw [g_lib.ERR_INVALID_PARAM, "Error writing to globus_token. No such user in u collection '" + user_id + "'"];
+                        }
+
+                        // find or insert collection
+                        const collection_search_key = other_token_data;
+                        let globus_collection = g_db.globus_coll.exists({_key: collection_search_key});
+                        if (!globus_collection) {
+                            // TODO: what other information should live on the globus_coll document?
```
t-ramz, line 537: I want a discussion here if possible. I'm not sure what we should be doing with the `globus_coll` as it is not yet defined in any place I could find.
We can make moves to store important data now, or pass that on to later - in which case I should remove the additional field.

core/database/foxx/tests/user_fixture.js
```diff
@@ -0,0 +1,24 @@
+const { db } = require("@arangodb");
+const g_lib = require("../api/support");
+
+const uid = "testUser0";
```
t-ramz, line 4: It would be nice to also export this and make it available to tests. I'm not sure how I want to do that right now, and I think it's out of scope. Just a thought.

core/database/foxx/api/user_router.js
```diff
@@ -516,9 +519,47 @@ router.get('/token/set', function(req, res) {
...
+                            globus_collection = g_db.globus_coll.save({_key: collection_search_key, name: "some name we could look up"});
+                        }
+
+                        const token_key = user_doc._key + "_" + globus_collection._key + "_" + token_type;    // TODO: key formatting/convention
```
t-ramz, line 541: What should we do for these token keys? I went with the mapping of u:globus_coll:type because I feel like that would make the most sense but I am likely not seeing everything.

core/database/foxx/api/user_router.js
```diff
@@ -516,9 +519,47 @@ router.get('/token/set', function(req, res) {
...
+                            _to: globus_collection._id,
+                            type: token_type,
+                            ...obj
+                        };
+
+                        console.log("writing to edge ", token_doc);
+                        const token_doc_upsert = g_db.globus_token.insert(
+                            token_doc,
+                            {
+                                overwriteMode: "replace"    // TODO: perhaps use 'update' and specify values for true upsert
```
t-ramz, line 554: For now we can probably kick this can down the road, but upserts would be helpful in my opinion.

core/database/foxx/api/user_router.js
```diff
@@ -516,9 +519,47 @@ router.get('/token/set', function(req, res) {
...
+                        // find or insert collection
+                        const collection_search_key = other_token_data;
+                        let globus_collection = g_db.globus_coll.exists({_key: collection_search_key});
+                        if (!globus_collection) {
+                            // TODO: what other information should live on the globus_coll document?
+                            globus_collection = g_db.globus_coll.save({_key: collection_search_key, name: "some name we could look up"});
+                        }
+
+                        const token_key = user_doc._key + "_" + globus_collection._key + "_" + token_type;    // TODO: key formatting/convention
+                        const token_doc = {
```
t-ramz, line 542: Are there any other important fields?

core/database/foxx/tests/user_router.test.js
```diff
@@ -0,0 +1,71 @@
...
+// Integration test of API
+const { expect } = require("chai");
+const request = require("@arangodb/request");
+const { baseUrl } = module.context;
+const { db } = require("@arangodb");
+
+const usr_base_url = `${baseUrl}/usr`
+
+
+// NOTE: describe block strings are compared against test specification during test call, not file name
```
t-ramz, line 13: 🙃

t-ramz (commented):
I have a few things I want some input on!

t-ramz:
We'll need to pull in the results of #1187 for fixes to CI

core/database/foxx/api/user_router.js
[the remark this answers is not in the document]
t-ramz: I'm not sure what is meant here. What collision would be prevented by "::" that would be introduced by underscores?

Regardless of the answer, I had considered and initially planned to utilize colons to represent the relationship that the key also represents, but [the arango docs](https://docs.arangodb.com/3.10/concepts/data-structure/documents/#user-specified-keys) suggest we do not use colons as part of the key because a few supported graph structures use the colon character as a separator in keys.

core/database/foxx/tests/user_router.test.js
```diff
@@ -0,0 +1,71 @@
...
+describe("user_router: the Foxx microservice user_router token/set endpoint", () => {
```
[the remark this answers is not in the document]
t-ramz: Full testing may be out of scope for this ticket, but this is a great suggestion nonetheless.

It may also be within scope and very helpful to both review error conditions introduced in this PR and test for those.

core/database/foxx/api/user_router.js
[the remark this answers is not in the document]
t-ramz: This is a great point, but I am not sure we will ever need to decompose the key found here. It is a composite of other items included in the document, and so if we want the user for example, we can access the `._from` field of the document.
I am still open to discussion on what the best choice may be, but I do not see this as a large issue as the code is now.

core/database/foxx/api/user_router.js
AronPerez: Trying to get clarification here, does the `globus_coll` doc have:
```javascript
    class GlobusCollection {
        +string _key
        +string name
    }
```

Maybe...
```javascript
description: "Brief description of the collection.",
owner: "User or entity that owns the collection.",
created_at: "Timestamp of when the collection was created.",
updated_at: "Timestamp of the last update to the collection.",
```

core/database/foxx/tests/user_fixture.js
AronPerez: Right now yes, that being said it will definately be on the agenda as QA when we're wrapping up this work.

core/database/foxx/api/user_router.js
```diff
@@ -516,9 +519,47 @@ router.get('/token/set', function(req, res) {
                         refresh: req.queryParams.refresh,
                         expiration: Math.floor(Date.now() / 1000) + req.queryParams.expires_in
                     };
-                    g_db._update(user_id, obj, {
-                        keepNull: false
-                    });
+                    if (token_type && other_token_data) {
+
+                        // NOTE: the edge is only defined for the `u` collection and the `globus_coll` collection.
```
t-ramz, line 524: Upon another look, it seems like `getUserFromClientID` will check against edge(s) to ensure there is a link between either a `uuid` or a domain account identity to a user account `u`, so this should either be replaced with simply using `user_id` above, or because we need to always have access to the key, `const user_doc = g_db.u.document(user_id);`

core/database/foxx/api/user_router.js
t-ramz: Currently, from what I can tell, `globus_coll` is totally unused. The work being done here is defining what the document's structure will look like, with a lot of room for change down the road.

I think you're right that these fields would be great adds, I am just unsure how I would get that additional information as of right now. Except for `created_at` and `updated_at` which I assume are metadata for the records themselves.

core/database/foxx/api/user_router.js
AronPerez: Here's my suggestion:
`globus_coll:type:u`


1. Put the `globus_collection.key` first because:
  * It's the primary scope/context for the token
  * Makes it easier to query all tokens for a specific collection
  * Aligns with how Globus structures their scopes (collection UUID first)
2. Put `token_type` second because:
  * Different token types may be needed for the same collection
  * Helps organize tokens by their purpose/scope
  * Makes it easy to query all tokens of a specific type for a collection                                                                                                                                                              
3. Put `user_doc.key` last because:
  * Multiple users may have tokens for the same collection/type combination
  * Follows principle of most general to most specific
  * Makes hierarchical organization clearer                                                                                                                                                                                            

This structure would make it easier to:

* Query all tokens for a collection
* Query all tokens of a specific type for a collection
* Query a specific user's token for a collection/type combination

The format would look like: `1cbaaee5-b938-4a4e-87a8-f1ec4d5d92f9:1:user123`

core/database/foxx/api/user_router.js
AronPerez: Possibly:

```javascript
{
    "scope": "urn:globus:auth:scope:transfer.api.globus.org:all", // Base scope
    "dependent_scopes": [], // Array of dependent collection-specific scopes
    "request_time": Math.floor(Date.now() / 1000), 
    "last_used": Math.floor(Date.now() / 1000),
    "status": "<active/inactive/expired>", // For tracking token validity
    "collection_display_name": globus_collection.name || null // For easier reference   
}

```

core/database/foxx/api/user_router.js
t-ramz: Ok! I like this suggestion. One of the main reasons I had user first was that the user was context we likely always have, but in a sense we can also use that fact to reduce options earlier in the search, and also moving from more general to more specific makes a lot of sense.

Thank you for the input, I'll get this order changed 🤠

core/database/foxx/api/user_router.js
t-ramz: Great suggestions! I think the display name and scopes will be harder to retrieve, but we can absolutely stub them for now and get the real content later!

core/database/foxx/api/user_router.js
```diff
@@ -516,9 +519,54 @@ router.get('/token/set', function(req, res) {
                         refresh: req.queryParams.refresh,
                         expiration: Math.floor(Date.now() / 1000) + req.queryParams.expires_in
                     };
-                    g_db._update(user_id, obj, {
-                        keepNull: false
-                    });
+                    if (token_type && other_token_data) {
+                        // We need the '_key' from the user document, which is not necessarily a part of the above checks.
+                        const user_doc = g_db.u.document(user_id);
```
JoshuaSBrown, line 524: I believe you should do this in the if else above, you already have the user document if it is the client, in that case you can avoid making another call to the database, you will still have to make the call if the user_id is the subject.

```
                    let user_doc = null;
                    if (req.queryParams.subject) {
                        user_id = req.queryParams.subject;
                        if (!g_db.u.exists(user_id))
                            throw [g_lib.ERR_INVALID_PARAM, "No such user '" + user_id + "'"];
                        g_lib.ensureAdminPermUser(client, user_id);
                        user_doc = g_db.u.document(user_id);
                    } else {
                        user_id = client._id;
                        user_doc = client;
                    }
```

core/database/foxx/api/user_router.js
JoshuaSBrown: @AronPerez  makes a good suggestion about update time and create time. The data_router also does this for records for consistency consider using the same language.

Below is from the data_router. I'm not really a fan of ct and ut, but its probably more important to be consistent.
```
    var time = Math.floor(Date.now() / 1000),
        obj = {
            size: 0,
            ct: time,
            ut: time,
            owner: owner_id,
            creator: client._id
        },
 ```
 
 In terms of other information, this is what I can think of. 
 
 ```
 _key: "This should be the uuid of the collection",
 _id: "globus_coll/<uuid of the collection",
 scopes: "",
 ca: Math.floor(Date.now() / 1000),
 ua: Math.floor(Date.now() / 1000), 
 type: "<mapped, guest>", - will need to know this because it effects consent flows. There may be other collection types, I'm not yet sure if it matters if it is a personal or server collection type, that also might be important.
 ha_enabled: "true/false"  - will also need to know this because of consent flows
 ```
 
This information should be available to you already from the web side when you get the access and refresh tokens.

File: core/database/foxx/api/user_router.js (hunk 5 above)
Comment (JoshuaSBrown) [reply core/database/foxx/api/user_router.js:544]: I would avoid including the collection_display_name in the edge document, as it will add a piece of information that might need to be synchronized with what is in the actual globus_coll document.

The scopes you need should already be available when you requested the token from the web server, if you needed any. A good default can be found by looking in the authz part of the web server.

https://github.com/ORNL/DataFed/blob/b6b555d4fdd2497bd30eb1ff3d328a47f3058f9a/web/datafed-ws.js#L202

File: core/database/tests/test_foxx.sh [hunk 10]
```diff
@@ -128,6 +128,14 @@ if ! command -v foxx > /dev/null 2>&1; then
...
 PATH_TO_PASSWD_FILE=${SOURCE}/database_temp.password
+
+# set up test fixtures, this script should be idempotent, this script is described in the manifest
+"${FOXX_PREFIX}foxx" script -u "${local_DATABASE_USER}" \
+   --server "tcp://${DATAFED_DATABASE_HOST}:8529" \
+    -p "${PATH_TO_PASSWD_FILE}" \
+    --database "${local_DATABASE_NAME}" \
+    "/api/${local_FOXX_MAJOR_API_VERSION}" user-fixture
+
 if [ "$TEST_TO_RUN" == "all" ]
```
JoshuaSBrown, line 139: CMake already provides a means for doing this take a look here.

https://github.com/ORNL/DataFed/blob/ebef1b003a5b270ae11023a0aadf8c73f0726404/core/database/CMakeLists.txt#L1C1-L29C8

If you want to only make these changes for this test. You can add a separte cmake setup fixture. Otherwise, you can add it directly in https://github.com/ORNL/DataFed/blob/1180-refactor-authz/core/database/tests/test_setup.sh, 

Best practice would probably be to keep it separate. And add something like this to the CMakeLists.txt file. You will need a separate shell script which should probably be placed in here. https://github.com/ORNL/DataFed/tree/1180-refactor-authz/core/database/tests

```
add_test(NAME foxx_setup_user COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/tests/test_setup_user.sh")
add_test(NAME foxx_user COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/tests/test_foxx.sh" -t "unit_user")

set_tests_properties(foxx_setup PROPERTIES FIXTURES_SETUP FoxxUser)
set_tests_properties(foxx_setup_user PROPERTIES FIXTURES_SETUP FoxxUser DEPENDS foxx_setup)
set_tests_properties(foxx_teardown PROPERTIES FIXTURES_CLEANUP FoxxUser)
set_tests_properties(foxx_user PROPERTIES FIXTURES_REQUIRED FoxxUser)
```

Might want to take a look at this documentation. 

https://cmake.org/cmake/help/latest/prop_test/FIXTURES_SETUP.html

/core/database/tests/test_setup_user.sh - this is incomplete
```
"${FOXX_PREFIX}foxx" script -u "${local_DATABASE_USER}" \
   --server "tcp://${DATAFED_DATABASE_HOST}:8529" \
    -p "${PATH_TO_PASSWD_FILE}" \
    --database "${local_DATABASE_NAME}" \
    "/api/${local_FOXX_MAJOR_API_VERSION}" user-fixture
```

core/database/foxx/tests/user_router.test.js
```diff
@@ -0,0 +1,130 @@
...
+    const test_param_indexes = [0,1,2,3,4];
+    const test_params = test_param_indexes.map((param_index) => {
+        return {
+            _key: "testUser" + param_index,
+            access: "access_token" + param_index,
+            refresh: "refresh_token" + param_index,
+        }
+    });
+    const test_edge_params = {
+        type: 4,  // transfer token
```
JoshuaSBrown, line 24: Where are these numbers defined? These should probably be an enum instead of a magic number, unless I'm missing something. They could be defined in the foxx/api/support.js file with the other enums.

core/database/foxx/api/user_router.js
```diff
@@ -530,6 +578,8 @@ router.get('/token/set', function(req, res) {
     .queryParam('access', joi.string().required(), "User access token")
     .queryParam('refresh', joi.string().required(), "User refresh token")
     .queryParam('expires_in', joi.number().integer().required(), "Access token expiration timestamp")
+    .queryParam('type', joi.number().optional(), "Type of token being stored")    // TODO: currently represented as underlying integer from SDMS::AccessTokenType, do we want strings?
```
JoshuaSBrown, line 581: I would make a comment where you define the Access Token Type in Foxx, class that it needs to be consistent with what is in here:

common/proto/common/SDMS.proto

We want enums. Great question. You can use the string validation with enums. 
Something like this: 
```
const AccessTokenType = { GENERIC: 1,
    GLOBUS: 2,
    GLOBUS_AUTH: 3,
    GLOBUS_TRANSFER: 4 };
```
```
'type', joi.integer().valid(...Object.values(AccessTokenType)),
```

core/database/foxx/tests/user_router.test.js
JoshuaSBrown: Nevermind, I now understand see earlier comment.

core/database/foxx/api/user_router.js
```diff
@@ -516,9 +519,54 @@ router.get('/token/set', function(req, res) {
...
                     };
-                    g_db._update(user_id, obj, {
-                        keepNull: false
-                    });
+                    if (token_type && other_token_data) {
+                        // We need the '_key' from the user document, which is not necessarily a part of the above checks.
+                        const user_doc = g_db.u.document(user_id);
+
+                        // find or insert collection
+                        const collection_search_key = other_token_data;
```
JoshuaSBrown, line 527: An important note here, you are assuming this just the UUID. Other token data is a string which is not guaranteed to just be the uuid.

You will need a function here to verify the contents. For instance, it may also contain the scopes.

You might ask why not include this as a separate query parameter? My thought behind this is if we support tokens to other applications or services they may not have a scope option. 

We need to have a discussion around what the other_token_data query parameter should and can contain, if it is a globus token. Right now, it should contain scopes and the uuid. What the format is also needs to be defined.

Perhaps?

"<uuid>:<scopes>"

You could then have a JavaScript function that parses the string. 
```
const { uuid, scopes } = parseGlobusTokenData( other_token_data );
```

JoshuaSBrown: If we go this way, we will have to also be sure the core service knows how to correctly pass the globus token data as well.

core/database/foxx/tests/user_router.test.js
```diff
@@ -0,0 +1,130 @@
...
+
+        const user_token_edge = db.globus_token.outEdges(user_token_data._id);
+        expect(user_token_edge).to.be.empty;
+    });
+
+    it("should update only a valid user's token when additionally provided other token data", () => {
+        // arrange
+        const lcoal_test_params = test_params[2];
+        const query_params = {
+            ...lcoal_test_params,
```
JoshuaSBrown, line 72: I think you have a typo here, this should be local not lcoal.

core/database/foxx/api/user_router.js
```diff
@@ -516,9 +519,54 @@ router.get('/token/set', function(req, res) {
                         refresh: req.queryParams.refresh,
                         expiration: Math.floor(Date.now() / 1000) + req.queryParams.expires_in
                     };
-                    g_db._update(user_id, obj, {
-                        keepNull: false
-                    });
+                    if (token_type && other_token_data) {
```
JoshuaSBrown, line 522: If one of this is defined and the other is not I would recommend throwing an error here because it indicates that the user doesn't realize the information they are providing will not be used unless both pieces of information are defined.

core/database/foxx/tests/user_router.test.js
```diff
@@ -0,0 +1,130 @@
...
+        const response = request.get(request_string);
+
+        // assert
+        expect(response.status).to.equal(204);
+
+        const user_token_data = db.u.document({_key: query_params._key});
+        expect(user_token_data).to.include(query_params);
+    });
+
+    it("should update only a valid user's token when additionally provided type", () => {
```
JoshuaSBrown, line 45: I think this should throw an error, see earlier comment. 

NOTE: if we actually stored type in the user document I think this test would be fine the way it is.

core/database/foxx/tests/user_router.test.js
```diff
@@ -0,0 +1,130 @@
...
+
+        const user_token_data = db.u.document({_key: query_params._key});
+        expect(user_token_data).to.include(local_test_params);
+
+        const user_token_edge = db.globus_token.outEdges(user_token_data._id);
+        expect(user_token_edge).to.be.empty;
+    });
+
+    it("should update only a valid user's token when additionally provided other token data", () => {
+        // arrange
```
JoshuaSBrown, line 69: This test should also throw an error because including additional information that is not actually being stored or used by the route indicates the user doesn't understand what the route is supposed to be doing. In this case other info is being used at all.

core/database/foxx/tests/user_router.test.js
```diff
@@ -0,0 +1,130 @@
...
+
+        const user_token_data = db.u.document({_key: query_params._key});
+        expect(user_token_data).to.include(lcoal_test_params);
+
+        const user_token_edge = db.globus_token.outEdges(user_token_data._id);
+        expect(user_token_edge).to.be.empty;
+    });
+
+    it("should accept a valid user's token with additional type and globus collection data and add it to globus_token edge", () => {
+        // arrange
```
JoshuaSBrown, line 92: This test is great!

I would also recommend adding tests that throw errors if non valid other_token_data is provided.

JoshuaSBrown (changes requested):
@t-ramz great start. 

Unfortunately, I think we have a bit more work to do before this can be merged. Some of the pieces can be placed in a different PR. The pieces that I think we need to address before this gets merged are.

1. The test fixtures, the approach you have used is bit redundant with options that are already available.
2. The use of update_time and create_time, make those consistent by using ut and ct. 
3. Finally, other_globus_info this needs to include the scope and a standaradized format needs to be defined, I've made a suggestion.

Everything, else I have mentioned I think can be addressed in a follow up PR.

JoshuaSBrown:
One more recommendation, when you get this cleaned up a little bit can you add some documentation here:

https://confluence.ccs.ornl.gov/pages/viewpage.action?pageId=344163314

core/database/foxx/api/user_router.js
t-ramz: Ah ok yeah that looks like it should work, I was unsure the `_document` method being used here https://github.com/ORNL/DataFed/blob/devel/core/database/foxx/api/support.js#L610-L612 was equivalent to the collection's `document` method, and I assumed it returned a casted object with the `_id` field. I can't find any documentation on `db._document` so I had to work under assumptions of worst case here.

core/database/foxx/api/user_router.js
t-ramz: Yeah the assumption here was just so I could get something functional, so this is a great callout. I will add a comment with an explanation that this assumption was made and the section of code will need updates.

What other application tokens would we accept here? Does Globus allow for usage of other authorization providers' tokens? Currently, the UUID is represented as tied to a Globus collection here, so it may not make a lot of sense. I could see it working if we make the `token_type` represent determine what logic we apply, but the majority of those options should maybe be stubbed for now.

So, for `GENERIC`, in the future, we parse it as some identifier, and no scopes in your example. For `GLOBUS_TRANSFER` we could parse it as it is now, and `GLOBUS_AUTH` or just plain `GLOBUS` we parse it as the code previously did. In this case, we would need to always pass at least one option.

core/database/foxx/api/user_router.js
t-ramz: Great callouts, I will update the document insert!

core/database/foxx/api/user_router.js
t-ramz: Ok yeah passing the scopes through the `other_token_data` should be doable, and I can make it represented at least here. I will also want to do a bit more verification, but that could be great.

core/database/foxx/api/user_router.js
t-ramz: Whoa, that's very cool! I will work on these 🫡

core/database/foxx/api/user_router.js
t-ramz: This will be fine and should be documented as to what the types mean, perhaps another type should be added for the default token using the code's previous logic. Name could be `GLOBUS_DEFAULT`

core/database/foxx/api/support.js
```diff
@@ -2837,5 +2845,32 @@ module.exports = (function() {
         }
     };
 
+    obj.parseOtherTokenData = (token_type, other_token_data) => { // TODO: type enforcement of passed data
+        let return_data = {};
+        // TODO: other token types
+        if (token_type === obj.AccessTokenType.GLOBUS_TRANSFER) {
+            // expect the data to be in the format of "<UUID>|<scopes>"
```
t-ramz, line 2852: I think this could and should be documented in a docstring, which also mostly addresses typing.

core/database/foxx/api/user_router.js
```diff
@@ -495,30 +495,81 @@ router.get('/token/set', function(req, res) {
...
                     console.log("updating tokens for", user_id, "acc:", req.queryParams.access, "exp:", req.queryParams.expires_in);
                     var obj = {
                         access: req.queryParams.access,
                         refresh: req.queryParams.refresh,
                         expiration: Math.floor(Date.now() / 1000) + req.queryParams.expires_in
                     };
-                    g_db._update(user_id, obj, {
-                        keepNull: false
-                    });
+                    if ((token_type || other_token_data) && !edge_update) {
```
t-ramz, line 525: Is the intent of this clear?

core/database/foxx/api/user_router.js
```diff
@@ -495,30 +495,81 @@ router.get('/token/set', function(req, res) {
...
+                        throw [g_lib.ERR_INVALID_PARAM, "Invalid parameters passed, type and other_token_data depend on one another."];
+                    }
+                    else if (edge_update && token_type === g_lib.AccessTokenType.GLOBUS_TRANSFER) {
+                        // find or insert collection
+                        const { uuid: collection_search_key, scopes } = g_lib.parseOtherTokenData(g_lib.AccessTokenType.GLOBUS_TRANSFER, other_token_data);
+                        let globus_collection = g_db.globus_coll.exists({_key: collection_search_key});
+                        if (!globus_collection) {
+                            globus_collection = g_db.globus_coll.save({
```
t-ramz, line 533: Have a few questions here for more eyes.

core/database/tests/test_foxx.sh
t-ramz: I went with the second route 👍

core/database/foxx/api/user_router.js
t-ramz: Ok so for the `type` and `ha_enabled` options, should I add those as expected fields on the `other_token_data` string?

core/database/foxx/api/user_router.js
t-ramz: I've addressed this by both adding a new piece to the enum, which will need to be updated in #1120 and #1127, and setting a default in the query params.

core/database/foxx/api/user_router.js
```diff
@@ -495,30 +495,98 @@ router.get('/token/set', function(req, res) {
...
                     var obj = {
                         access: req.queryParams.access,
                         refresh: req.queryParams.refresh,
                         expiration: Math.floor(Date.now() / 1000) + req.queryParams.expires_in
                     };
-                    g_db._update(user_id, obj, {
-                        keepNull: false
-                    });
+
+                    if (Boolean(other_token_data) && token_type === g_lib.AccessTokenType.GLOBUS_DEFAULT) {
```
t-ramz, line 525: These validations can occur much earlier in the function to save some cycles and allocations.

core/database/foxx/tests/user_router.test.js
t-ramz: I made it use the enum.

core/database/foxx/api/user_router.js
```diff
@@ -495,30 +495,98 @@ router.get('/token/set', function(req, res) {
...
+                        }
+                        case g_lib.AccessTokenType.GLOBUS_AUTH: {
+                            break;
+                        }
+                        case g_lib.AccessTokenType.GLOBUS_TRANSFER: {
+                            // find or insert collection
+                            const { uuid: collection_search_key, scopes } = g_lib.parseOtherTokenData(g_lib.AccessTokenType.GLOBUS_TRANSFER, other_token_data);
+                            let globus_collection = g_db.globus_coll.exists({_key: collection_search_key});
+                            if (!globus_collection) {
+                                globus_collection = g_db.globus_coll.save({
```
t-ramz, line 547: I have some more questions here

core/database/foxx/api/user_router.js
t-ramz: Duplicated in new changes

core/database/foxx/api/user_router.js
```diff
@@ -571,9 +594,71 @@ router
...
+                                    _key: collection_search_key,
+                                    name: "Newly Inserted Collection",
+                                    description: "The collection description",
+                                    required_scopes: scopes, // string - Is this always the minimum required set?
+                                    owner: "", // Should this be looked up on Globus?
+                                    ct: Math.floor(Date.now() / 1000),
+                                    ut: Math.floor(Date.now() / 1000),
+                                    type: "mapped/guest", // How do we determine this from here? Should it be passed as part of the params?
```
JoshuaSBrown, line 629: You cannot determine this here, similar to scopes this will have to be passed in. As this is not a trivial amount of work I would create an issue to address it in a follow up PR. For now you can just assume it is a mapped collection.

core/database/foxx/api/user_router.js
```diff
@@ -571,9 +594,71 @@ router
...
+                                    ha_enabled: false, // boolean - How do we determine this from here? Should it be passed as part of the params?
```
JoshuaSBrown, line 630: For now just assume ha is turned off, I would do the same as the above comment. This can be addressed in a separate PR and issue.

core/database/foxx/api/user_router.js
JoshuaSBrown: The only reason a Guest collection token would need to be saved is if HA is turned on and a consent flow is still required.

core/database/foxx/api/user_router.js
```diff
@@ -571,9 +594,71 @@ router
...
+                                globus_collection._key + "_" + token_type + "_" + user_doc._key;
+                            const token_doc = {
+                                _key: token_key,
+                                _from: user_id, // the uid field
+                                _to: globus_collection._id,
+                                type: token_type,
+                                dependent_scopes: scopes,
+                                request_time: Math.floor(Date.now() / 1000),
+                                last_used: Math.floor(Date.now() / 1000),
+                                status: "active", // TODO: are there other checks we should have right away?
```
JoshuaSBrown, line 644: Is this for understanding if the token is active? I'm a bit confused where are you saving the refresh and access tokens?

I'm assuming that it would only refer to whether the access token is still valid. If the refresh token becomes invalid it should be removed from the database. 

Any call that is using the tokens should check if the access token is valid and refresh it automatically if it isn't. The only place this would likely be used is between calls. I.e. you check if the token is valid and discover it is not so you change this to inactive, but then immediately use the refresh token and and update it. 

There are two ways I know of that the token becomes invalid:

1. Revocation
2. Expiration

A possible token usage workflow would be

## Flow 1 - invalid refresh and access
1. Get token
2. Check if expired
3. If not try to use token
4. If error try to use refresh token
5. if error
6. Will need to notify the user their refresh token is invalid, will also likely need to remove the token edge document

## Flow 2 - invalid access
1. Get token
2. Check if expired
3. If not try to use token
4. If error try to use refresh token
5. Reset access token 
6. Make request

core/database/foxx/api/user_router.js
```diff
@@ -571,9 +594,71 @@ router
...
+                            const token_doc_upsert = g_db.globus_token.insert(token_doc, {
+                                overwriteMode: "replace", // TODO: perhaps use 'update' and specify values for true upsert
+                            });
+                            break;
+                        }
+                        case g_lib.AccessTokenType.GLOBUS_DEFAULT: {
+                            // Existing logic, default actions
+                            g_db._update(user_id, obj, {
+                                keepNull: false,
+                            });
```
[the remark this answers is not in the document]
t-ramz: Line 592, and I greatly dislike the naming so I would be happy to change it.

core/database/foxx/api/user_router.js
t-ramz: I did some verification and it looks to be easily passable from the web server all the way to the database!

core/database/foxx/api/user_router.js
JoshuaSBrown: You are talking about before you actually set it? It wouldn't really hurt anything to have a satement that checks the expiration and if it is expired make sure that the status is set to inactive.

core/database/foxx/api/user_router.js
```diff
@@ -571,9 +594,71 @@ router
...
+                            // GLOBUS_TRANSFER parse currently assumes uuid and scopes exist
+                            let globus_collection = g_db.globus_coll.exists({
+                                _key: collection_search_key,
+                            });
+                            if (!globus_collection) {
+                                globus_collection = g_db.globus_coll.save({
+                                    _key: collection_search_key,
+                                    name: "Newly Inserted Collection",
+                                    description: "The collection description",
+                                    required_scopes: scopes, // string - Is this always the minimum required set?
```
JoshuaSBrown, line 625: Are you wanting to add additional checks, is that where your question is coming from?

When you say 'set' are you referring to just the scopes or the informatino packaged with the token.?

core/database/foxx/api/user_router.js
```diff
@@ -571,9 +594,71 @@ router
...
+                                    owner: "", // Should this be looked up on Globus?
```
JoshuaSBrown, line 626: Yes, all of this information will likely come from Globus with a collection get call to the API. Again I wouldn't worry about collecting it for this PR.

core/database/foxx/api/user_router.js
```diff
@@ -571,9 +594,71 @@ router
...
+                                globus_collection._key + "_" + token_type + "_" + user_doc._key;
+                            const token_doc = {
+                                _key: token_key,
+                                _from: user_id, // the uid field
+                                _to: globus_collection._id,
+                                type: token_type,
+                                dependent_scopes: scopes,
+                                request_time: Math.floor(Date.now() / 1000),
+                                last_used: Math.floor(Date.now() / 1000),
+                                status: obj.expiration > Math.floor(Date.now() / 1000) ? "active" : "inactive",
```
t-ramz, line 644: Since the expiration field is set within this function as well, it may make sense to add a padded window as well. This is especially true if other areas in code already have a window where they will go ahead and fetch a refresh e.g. `if ( token.expiration < now() + 100 //seconds ) { refreshToken(); }`

core/database/CMakeLists.txt
```diff
@@ -13,11 +13,15 @@ configure_file(
...
   add_test(NAME foxx_setup COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/tests/test_setup.sh")
   add_test(NAME foxx_teardown COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/tests/test_teardown.sh")
+  add_test(NAME foxx_db_fixtures COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/tests/test_fixture_setup.sh")
   add_test(NAME foxx_version COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/tests/test_foxx.sh" -t "unit_version")
   add_test(NAME foxx_support COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/tests/test_foxx.sh" -t "unit_support")
+  add_test(NAME foxx_user_router COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/tests/test_foxx.sh" -t "user_router")
 
   set_tests_properties(foxx_setup PROPERTIES FIXTURES_SETUP Foxx)
   set_tests_properties(foxx_teardown PROPERTIES FIXTURES_CLEANUP Foxx)
+  set_tests_properties(foxx_db_fixtures PROPERTIES FIXTURES_SETUP FoxxDBFixtures FIXTURES_REQUIRED Foxx)
```
JoshuaSBrown, line 23: Great!

core/database/foxx/api/user_router.js
```diff
@@ -571,9 +594,71 @@ router
...
+                            });
+                            if (!globus_collection) {
+                                globus_collection = g_db.globus_coll.save({
+                                    _key: collection_search_key,
+                                    name: "Newly Inserted Collection",
+                                    description: "The collection description",
+                                    required_scopes: scopes, // string - Is this always the minimum required set?
+                                    owner: "", // Should this be looked up on Globus?
+                                    ct: Math.floor(Date.now() / 1000),
+                                    ut: Math.floor(Date.now() / 1000),
```
JoshuaSBrown, line 628: Great!

JoshuaSBrown (approved):
Ok, I'm happy with the state of the PR. I would recommend opening up issues that require better definition. You have added several TODO comments that need additional exploration but are not immediate priorities.

AronPerez (approved):

PR #1221, opened 2025-01-09
1219 daps bug gitlab ci gcs log

JoshuaSBrown:
# PR Description

* Using filter with ancestor can return more than one container name, this causes problems when running with docker logs command which expects a single argument.
* Bug was also found to be a difference in container names when building the images vs when testing them and calling the logs. Container names when building where called gcs and when testing were called gcs-authz. 

# Tasks

## Summary by Sourcery

CI:
- Fix a bug in the log command where filtering by ancestor could return multiple container names.

### DIFF
--- a/.gitlab/common.yml
+++ b/.gitlab/common.yml
@@ -301,7 +301,7 @@
     - docker ps
     - export ANCESTOR_CONTAINERS=$(docker ps -a --filter "ancestor=${FULL_IMAGE_NAME}" --format "{{.Names}}") 
     - echo "Containers are: $ANCESTOR_CONTAINERS"
-    - echo "Matching against: ${COMPONENT}-${BRANCH_LOWER}"
+      #    - echo "Matching against: ${COMPONENT}-${BRANCH_LOWER}"
       #    - export MATCHING_CONTAINERS=$( echo "$ANCESTOR_CONTAINERS" | grep "${COMPONENT}-${BRANCH_LOWER}" || echo "")
       #    - echo "Matching containers are: $MATCHING_CONTAINERS" 
       #    - if [ -z "$MATCHING_CONTAINERS" ]; then
--- a/.gitlab/common.yml
+++ b/.gitlab/common.yml
@@ -301,10 +301,10 @@
     - docker ps
     - export ANCESTOR_CONTAINERS=$(docker ps -a --filter "ancestor=${FULL_IMAGE_NAME}" --format "{{.Names}}") 
     - VAR="Test"
-    - echo "This is failing: ${VAR}"
-    - echo "Matching against: ${COMPONENT}-${BRANCH_LOWER}"
+    - echo "This is failing ${VAR}"
+    - echo "Matching against ${COMPONENT}-${BRANCH_LOWER}"
     - export MATCHING_CONTAINERS=$( echo "$ANCESTOR_CONTAINERS" | grep "${COMPONENT}-${BRANCH_LOWER}" || echo "")
-    - echo "Matching containers are: $MATCHING_CONTAINERS" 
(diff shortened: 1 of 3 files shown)


.gitlab/common.yml
```diff
@@ -287,7 +287,26 @@
   stage: log
   script:
     - BRANCH_LOWER=$(echo "$CI_COMMIT_REF_NAME" | tr '[:upper:]' '[:lower:]')
-    - docker logs $(docker ps -a --filter "ancestor=${REGISTRY}/${PROJECT}/${COMPONENT}-${BRANCH_LOWER}" --format "{{.Names}}")
+    - echo "Lower branch is $BRANCH_LOWER"
+    - export FULL_IMAGE_NAME="${REGISTRY}/${PROJECT}/${COMPONENT}-${BRANCH_LOWER}"
+    - echo "Full image name is $FULL_IMAGE_NAME"
+    - export ANCESTORS=$(docker ps -a --filter "ancestor=${FULL_IMAGE_NAME}" --format "{{.Names}}")
+    - echo "Found Ancestors"
+    - echo "$ANCESTORS"
```
AronPerez, line 295: Are we using `ANCESTORS` anywhere else? The fix looks like `ANCESTOR_CONTAINERS is all we need.

.gitlab/common.yml
```diff
@@ -287,7 +287,26 @@
   stage: log
   script:
     - BRANCH_LOWER=$(echo "$CI_COMMIT_REF_NAME" | tr '[:upper:]' '[:lower:]')
-    - docker logs $(docker ps -a --filter "ancestor=${REGISTRY}/${PROJECT}/${COMPONENT}-${BRANCH_LOWER}" --format "{{.Names}}")
+    - echo "Lower branch is $BRANCH_LOWER"
```
AronPerez, line 290: Nit: We'll know what the `BRANCH_LOWER` from:
```yml
echo "Full image name is $FULL_IMAGE_NAME"
```

.gitlab/end_to_end.yml
```diff
@@ -312,13 +312,13 @@ end-to-end-gcs-authz-setup:
     - chown gitlab-runner "$HOST_LOG_FILE_PATH"
     - ./scripts/generate_datafed.sh
     - docker login "${REGISTRY}" -u "${HARBOR_USER}" -p "${HARBOR_DATAFED_GITLAB_CI_REGISTRY_TOKEN}"
-    - ./scripts/container_stop.sh -n "gcs-authz" -p
+    - ./scripts/container_stop.sh -n "${COMPONENT}" -p
```
AronPerez, line 315: Does `gcs-authz` need to be updated antwhere else?

.gitlab/stage_unit.yml
```diff
@@ -94,11 +94,11 @@ run-authz-unit-job:
     - chown gitlab-runner "$HOST_LOG_FILE_PATH"
     - ./scripts/generate_datafed.sh
     - docker login "${REGISTRY}" -u "${HARBOR_USER}" -p "${HARBOR_DATAFED_GITLAB_CI_REGISTRY_TOKEN}"
-    - ./scripts/container_stop.sh -n "gcs-authz" -p
+    - ./scripts/container_stop.sh -n "${COMPONENT}" -p
```
AronPerez, line 97: Same ❓  as above

AronPerez (changes requested):
* Questions about if exports are needed
* If we need to update a variable usage anywhere else

.gitlab/end_to_end.yml
JoshuaSBrown: No this was the reason for the source of the error. It is already used correctly in the build files. The differnece between the build and the tests running is caused by this.

AronPerez (approved):

PR #1227, opened 2025-01-13
Update nlhoman json version

JoshuaSBrown:
# PR Description

Unit testing in github action failes with using ubuntu latest because it is no longer compatible with 20.04. 

# Tasks

## Summary by Sourcery

Build:
- Update nlohmann json from 3.10.2 to 3.11.3.


.github/workflows/unit-tests.yml
```diff
@@ -2,7 +2,7 @@ name: Unit-Testing
 on: push
 jobs:
   unit-test:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
```
t-ramz, line 5: Is there a reason we're only pinning the deployment OS and not updating the dependency?

t-ramz (approved):
Seems fine, but I have a question 🤔

.github/workflows/unit-tests.yml
JoshuaSBrown: Other dependencies will likely be changed in the newer OS. It would make sense to pin the compilers but this is a quick fix and that would take more time to figure out.

PR #1234, opened 2025-01-15
Fix bug on task_router.js abort function

t-ramz:
# PR Description
Fix bug in `task/abort` Foxx path

# Tasks

## Summary by Sourcery

Fix incorrect validation of the "task/abort" Foxx route parameters.

Bug Fixes:
- Corrected the validation of the request body for the "task/abort" route to accept optional objects rather than only strings.

Enhancements:
- Improved the flexibility of the "task/abort" route by allowing more diverse parameter structures.

### DIFF
--- a/core/database/foxx/api/task_router.js
+++ b/core/database/foxx/api/task_router.js
@@ -246,7 +246,7 @@ router
         }
     })
     .queryParam("task_id", joi.string().required(), "Task ID")
-    .body(joi.string().required(), "Parameters")
+    .body(joi.object().optional(), "Parameters")
     .summary("Abort a schedule task")
     .description("Abort a schedule task and return list of new runnable tasks.");
 


core/database/foxx/api/task_router.js
```diff
@@ -246,7 +246,7 @@ router
         }
     })
     .queryParam("task_id", joi.string().required(), "Task ID")
-    .body(joi.string().required(), "Parameters")
+    .body(joi.object().optional(), "Parameters")
```
t-ramz, line 249: The code still isn't totally working, this is just a small fix because there was no way to reach the code 👍

AronPerez: Gets called here but that's about it https://github.com/ORNL/DataFed/blob/de98a85b7108fba1650a6a82f3757abffcf56ce7/core/server/DatabaseAPI.cpp#L3103

JoshuaSBrown (approved):
Clarification. The task abort endpoint is not working and is not used anywhere in the code. when Tony tried to use it, it was broken, the change does not affect any in use workflows because nothing calls it outside of the DatabaseAPI.cpp function which is not called.

AronPerez (approved):

core/database/foxx/api/task_router.js
t-ramz: It's never called. The containing function is only ever defined.

PR #1252, opened 2025-01-23
Fix config access due to race condition

JoshuaSBrown:
# PR Description

# Tasks

### DIFF
--- a/repository/gridftp/globus5/authz/source/AuthzWorker.cpp
+++ b/repository/gridftp/globus5/authz/source/AuthzWorker.cpp
@@ -54,7 +54,7 @@ std::string randomAlphaNumericCode() {
 
 namespace SDMS {
 
-AuthzWorker::AuthzWorker(struct Config *a_config, LogContext log_context)
+AuthzWorker::AuthzWorker(struct Config a_config, LogContext log_context)
     : m_config(a_config) {
 
   m_log_context = log_context;
@@ -62,9 +62,9 @@ AuthzWorker::AuthzWorker(struct Config *a_config, LogContext log_context)
   m_log_context.thread_id = 0;
 
   // Convert config item to string for easier manipulation
-  m_local_globus_path_root = std::string(m_config->globus_collection_path);
+  m_local_globus_path_root = std::string(m_config.globus_collection_path);
 
-  m_test_path = std::string(m_config->test_path);
+  m_test_path = std::string(m_config.test_path);
   // NOTE the test_path MUST end with '/' this is to prevent authorization
   // by accident to a subfolder i.e.
   //
@@ -97,9 +97,9 @@ AuthzWorker::AuthzWorker(struct Config *a_config, LogContext log_context)
  **/
 void AuthzWorker::initCommunicator() {
 
-  m_cred_options[CredentialType::PUBLIC_KEY] = m_config->pub_key;
-  m_cred_options[CredentialType::PRIVATE_KEY] = m_config->priv_key;
-  m_cred_options[CredentialType::SERVER_KEY] = m_config->server_key;
+  m_cred_options[CredentialType::PUBLIC_KEY] = m_config.pub_key;
+  m_cred_options[CredentialType::PRIVATE_KEY] = m_config.priv_key;
+  m_cred_options[CredentialType::SERVER_KEY] = m_config.server_key;
   CredentialFactory cred_factory;
   m_sec_ctx = cred_factory.create(ProtocolType::ZQTP, m_cred_options);
 
@@ -144,7 +144,7 @@ void AuthzWorker::initCommunicator() {
 
     return comm_factory.create(socket_options, credentials, timeout_on_receive,
                                timeout_on_poll);
-  }(authz_thread_id, m_config->server_addr, *m_sec_ctx);
+  }(authz_thread_id, m_config.server_addr, *m_sec_ctx);
 }
 
 /**
@@ -182,7 +182,7 @@ bool AuthzWorker::isTestPath(const std::string &posix_path) const {
   if (m_test_path.size() > 0 &&
       posix_path.compare(0, m_test_path.size(), m_test_path) == 0) {
     DL_INFO(m_log_context,
-            "Allowing request within TEST PATH: " << m_config->test_path);
+            "Allowing request within TEST PATH: " << m_config.test_path);
     return true;
   }
   return false;
@@ -574,7 +574,7 @@ int AuthzWorker::checkAuth(char *client_id, char *path, char *action) {
 
   auto auth_req = std::make_unique<Auth::RepoAuthzRequest>();
 
-  auth_req->set_repo(m_config->repo_id);
+  auth_req->set_repo(m_config.repo_id);
   auth_req->set_client(client_id);
   auth_req->set_file(sanitized_path);
   auth_req->set_action(action);
@@ -629,7 +629,7 @@ const char *getReleaseVersion() {
 
 // The same
 int checkAuthorization(char *client_id, char *object, char *action,
-                       struct Config *config) {
+                       struct Config config) {
 #if defined(DONT_USE_SYSLOG)
   SDMS::global_logger.setSysLog(false);
 #else
@@ -637,7 +637,7 @@ int checkAuthorization(char *client_id, char *object, char *action,
 #endif
   SDMS::global_logger.setLevel(SDMS::LogLevel::INFO);
   SDMS::global_logger.addStream(std::cerr);
-  auto log_path_authz = std::string(config->log_path);
+  auto log_path_authz = std::string(config.log_path);
   if (log_path_authz.length() > 0) {
     // Append to the existing path because we don't want the C++ and C code
     // trying to write to the same file
--- a/repository/gridftp/globus5/authz/source/Config.c
+++ b/repository/gridftp/globus5/authz/source/Config.c
@@ -1,0 +1,445 @@
+
+// Local private includes
+#include "AuthzLog.h"
+#include "Config.h"
+
+// Standard includes
+#include <ctype.h>
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdio.h>
+#include <string.h>
+
+// IMPORTANT: The DATAFED_AUTHZ_CFG_FILE env variable must be set in the gridFTP
+// service script (usually /etc/init.d/globus-gridftp-server). This variable
+// points to the configuration file used for DataFed comm settings
+
+/******************************************************************************
+ * Global Vars
+ ******************************************************************************/
+// Define the global configuration object
+static struct Config g_config;
+// Boolean value is used to control initialization of the global config structure
+// false means it has not yet been initialized and true means it has already been
+// initialized
+static bool config_loaded = false;
+
+// Initialize the read-write lock for the configuration object
+pthread_rwlock_t config_rwlock = PTHREAD_RWLOCK_INITIALIZER;
+
+/******************************************************************************
+ * File Scoped Functions
+ ******************************************************************************/
+
+/**
+ * @brief Sets a value to a destination buffer with length checks.
+ *
+ * This function copies a source string (`a_src`) to a destination buffer (`a_dest`) 
+ * while ensuring the length of the source string does not exceed the maximum allowed length 
+ * (`a_max_len`). If the source string is empty or too long, an error is logged, 
+ * and the function returns `true`. Otherwise, the source string is copied to the destination.
+ *
+ * @param a_label The label associated with the configuration value, used in error logs.
+ * @param a_dest A pointer to the destination buffer where the source string will be copied.
+ * @param a_src A pointer to the source string to be copied to the destination buffer.
+ * @param a_max_len The maximum allowable length for the destination buffer.
+ *
+ * @return `true` if an error occurred (e.g., empty source or source too long), `false` otherwise.
+ *
+ * @note The function uses `strncpy` to copy the source string into the destination buffer,
+ *       ensuring that no more than `a_max_len` characters are copied. If the source string 
+ *       is longer than the allowed maximum length, an error is logged, and the value is not copied.
+ */
+bool setConfigValInternal(const char *a_label, char *a_dest, const char *a_src,
+                  size_t a_max_len) {
+  size_t len = strlen(a_src);
+  if (len == 0) {
+    AUTHZ_LOG_ERROR("DataFed - '%s' value not set.\n", a_label);
+    return true;
+  }
+
+  if (len > a_max_len) {
+    AUTHZ_LOG_ERROR(
+        "DataFed - '%s' value too long (max %zu).\n",
+        a_label, a_max_len);
+    return true;
+  }
+
+  strncpy(a_dest, a_src, a_max_len);
+
+  return false;
+}
+
+/**
+ * @brief Loads a key from a file into a destination buffer.
+ *
+ * This function attempts to open a file specified by `a_filename` and reads the first line
+ * into the provided destination buffer (`a_dest`). It checks for errors when opening the file
+ * and reading the key. If successful, the key is stored in `a_dest` with any trailing
+ * carriage return or newline characters stripped.
+ *
+ * @param a_dest A pointer to the buffer where the key will be loaded.
+ * @param a_filename The name of the file from which the key is to be loaded.
+ *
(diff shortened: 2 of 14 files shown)


JoshuaSBrown:
@par-hermes format

repository/gridftp/globus5/authz/source/Config.c
```diff
@@ -0,0 +1,497 @@
...
+    err = setConfigValInternal(a_label, g_config.globus_collection_path, a_src,
+                               MAX_PATH_LEN);
+  } else {
+    pthread_rwlock_unlock(&config_rwlock);
+    return err;
+  }
+
+  // Release the write lock
+  pthread_rwlock_unlock(&config_rwlock);
+  return err;
```
AronPerez, line 441: Do we need the else? Seems like we'd release the write lock and return error anyhow.

repository/gridftp/globus5/authz/source/Config.c
```diff
@@ -0,0 +1,497 @@
...
+    line++;
+  }
+
+  // Skip empty lines or comments
+  if (*line == '\0' || *line == '#') {
+    return true; // Not an error; simply ignore
+  }
+
+  // Parse the line for key=value
+  if (sscanf(line, "%255[^=]=%767s", key, value) != 2) {
```
t-ramz, line 205: I'm a little confused by the format string, but it certainly seems to work.

repository/gridftp/globus5/authz/source/Config.c
```diff
@@ -0,0 +1,497 @@
...
+ *
+ * @note The missing configuration keys, if any, are logged in an error message.
+ * The check is performed by looking at the first character of each key in the
+ * `g_config` structure, and if it is a null character (`\0`), the field is
+ * considered missing.
+ *
+ * @note Assumes read write thread lock is in place.
+ */
+bool validateConfig() {
+  char missing[1024] = {0};
```
AronPerez, line 313: Does 1024 refer to some data size?

repository/gridftp/globus5/authz/source/Config.c
```diff
@@ -0,0 +1,497 @@
...
+ * parse the configuration line. It uses `setConfigValInternal` to set values
+ * for various configuration fields, and `loadKeyFile` for loading
+ * public/private keys from files. If the configuration key is `timeout`, the
+ * value is converted from a string to an integer using `atoi`. Any unrecognized
+ * configuration keys will result in an error message.
+ *
+ * @note Assumes read write thread lock is in place.
+ */
+bool parseConfigLine(const char *line, int line_number) {
+  char key[256], value[768];
```
AronPerez, line 244: 256 corresponds to the max id length correct? Seems like a [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)) that could benefit from an explicit constant name for readability purposes.

Same applies to value value.

repository/gridftp/globus5/authz/source/Config.c
```diff
@@ -0,0 +1,497 @@
...
+
+void allowConfigReinitialization() {
+  pthread_rwlock_wrlock(&config_rwlock);
+  config_loaded = false;
+  pthread_rwlock_unlock(&config_rwlock);
+}
+
+struct Config createLocalConfigCopy() {
+  pthread_rwlock_rdlock(&config_rwlock);
+  struct Config temp = g_config;
```
t-ramz, line 361: My C is rusty (ha!) and I've been looking at too much JS and Python is assignment like this `Copy` by default?

repository/gridftp/globus5/authz/source/Config.c
```diff
@@ -0,0 +1,497 @@
...
+  pthread_rwlock_unlock(&config_rwlock);
+  return temp;
+}
+
+// Function to safely get a configuration value
+bool getConfigVal(const char *a_label, char *a_dest, size_t a_max_len) {
+
+  // Acquire a read lock (allow concurrent reads)
+  pthread_rwlock_rdlock(&config_rwlock);
```
t-ramz, line 370: Nice!!!

repository/gridftp/globus5/authz/source/Config.c
```diff
@@ -0,0 +1,497 @@
...
+
+  bool err = false;
+  // Logic to get specific configuration value
+  if (strcmp(a_label, "repo_id") == 0) {
+    strncpy(a_dest, g_config.repo_id, a_max_len);
```
t-ramz, line 375: See my other comment about sizes on copy

repository/gridftp/globus5/authz/source/Config.c
```diff
@@ -0,0 +1,497 @@
...
+  } else if (strcmp(a_label, "user") == 0) {
+    strncpy(a_dest, g_config.user, a_max_len);
+  } else if (strcmp(a_label, "test_path") == 0) {
+    strncpy(a_dest, g_config.test_path, a_max_len);
+  } else if (strcmp(a_label, "log_path") == 0) {
+    strncpy(a_dest, g_config.log_path, a_max_len);
+  } else if (strcmp(a_label, "globus_collection_path") == 0) {
+    strncpy(a_dest, g_config.globus_collection_path, a_max_len);
+  } else {
+    pthread_rwlock_unlock(&config_rwlock);
```
t-ramz, line 393: Looks like we're set up for a single return, and this else can skip the unlock, write to dest and set `err=true;`

repository/gridftp/globus5/authz/source/Config.c
```diff
@@ -0,0 +1,497 @@
...
+    err = setConfigValInternal(a_label, g_config.user, a_src, MAX_ID_LEN);
+  } else if (strcmp(a_label, "test_path") == 0) {
+    err =
+        setConfigValInternal(a_label, g_config.test_path, a_src, MAX_PATH_LEN);
+  } else if (strcmp(a_label, "log_path") == 0) {
+    err = setConfigValInternal(a_label, g_config.log_path, a_src, MAX_PATH_LEN);
+  } else if (strcmp(a_label, "globus_collection_path") == 0) {
+    err = setConfigValInternal(a_label, g_config.globus_collection_path, a_src,
+                               MAX_PATH_LEN);
+  } else {
```
t-ramz, line 434: The else should likely set err to true

t-ramz (commented):
I have a few comments about `Config.c`

repository/gridftp/globus5/authz/source/Config.c
JoshuaSBrown: Good catch.

repository/gridftp/globus5/authz/source/Config.c
```diff
@@ -0,0 +1,497 @@
...
+    AUTHZ_LOG_ERROR("DataFed - '%s' value not set.\n", a_label);
+    return true;
+  }
+
+  if (len > a_max_len) {
+    AUTHZ_LOG_ERROR("DataFed - '%s' value too long (max %zu).\n", a_label,
+                    a_max_len);
+    return true;
+  }
+
```
AronPerez, line 74: General C question from this.

```c
 * This function copies a source string (`a_src`) to a destination buffer
 * (`a_dest`) while ensuring the length of the source string does not exceed the
 * maximum allowed length
```

Is it bad practice to ensure null termination when using  `strncpy`, even if we know we don't exactly need it? There's no scenario where `a_src` is longer than `a_max_len` cause
```c
if (len > a_max_len)
```

So no buffer overruns or memory leaks.

I was reading this https://cboard.cprogramming.com/c-programming/117061-null-termination-str-n-cpy-strlen.html#post872229

repository/gridftp/globus5/authz/source/Config.c
```diff
@@ -0,0 +1,497 @@
...
+    }
+  }
+
+  fclose(config_file);
+
+  if (!validateConfig()) {
+    AUTHZ_LOG_ERROR("Configuration validation failed.");
+    // Avoid trying to reload
+    config_loaded = true;
+    logRelease();
```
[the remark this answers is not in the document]
t-ramz: Good catch! Yes, early return will need to be unlocked.

repository/gridftp/globus5/authz/source/Config.c
```diff
@@ -0,0 +1,497 @@
...
+    return true;
+  }
+
+  if (len > a_max_len) {
+    AUTHZ_LOG_ERROR("DataFed - '%s' value too long (max %zu).\n", a_label,
+                    a_max_len);
+    return true;
+  }
+
+  strncpy(a_dest, a_src, a_max_len);
```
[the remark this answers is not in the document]
AronPerez: https://github.com/ORNL/DataFed/pull/1252#discussion_r1927718398

Will get stopped by this as far as I can tell.
```c
len > a_max_len
```

repository/gridftp/globus5/authz/source/Config.c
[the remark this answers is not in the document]
t-ramz: I think this function can, and should, be refactored to a single return for this reason. Maybe along the lines of

```C
bool err = false;
if (!config_loaded) {
    if (config_file) {
        while () ...
    } else {
        err = true;
    }
}
...
unlock();
return err;
```
Will also need to set `config_loaded` accordingly

repository/gridftp/globus5/authz/source/Config.c
JoshuaSBrown: Correct C is copy by default. You have to use & or a pointer for referneces.

repository/gridftp/globus5/authz/source/Config.c
JoshuaSBrown: If err is initialized as true, you are correct we can get rid of the else good point.

repository/gridftp/globus5/authz/source/Config.c
```diff
@@ -0,0 +1,497 @@
...
+}
+
+bool initializeGlobalConfig() {
+
+  // Acquire a write lock to modify configuration values
+  pthread_rwlock_wrlock(&config_rwlock);
+
+  // Initialized only at start
+  if (config_loaded) {
+    AUTHZ_LOG_INFO("Config file already loaded. Skipping reload.\n");
```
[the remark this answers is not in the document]
JoshuaSBrown: Looks to be missing an unlock good catch.

repository/gridftp/globus5/authz/source/Config.c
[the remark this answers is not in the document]
JoshuaSBrown: Good catches @AronPerez

repository/gridftp/globus5/authz/source/Config.c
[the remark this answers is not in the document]
t-ramz: That will prevent copying larger data into a smaller string, I am worried about the copy function trying to copy more than the source string has to offer.
e.g.
```C
size_t max = 15;
char* source = "hi";
char destination[max];
strncpy(destination, source, max);
```
where the first two characters of source are valid, but what about the next 13 in this case? It could try to copy that arbitrary memory, I'm not sure

repository/gridftp/globus5/authz/source/Config.c
JoshuaSBrown: This is a really good catch, and demonstrates your attention to detail. You are correct, strlen only returns the number of chars excluding the null pointer. This should raise an error if len + 1 > a_max_len. strncpy will auto append \0 as long as there is room.

repository/gridftp/globus5/authz/source/Config.c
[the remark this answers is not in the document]
JoshuaSBrown: When you use "" C auto appends the null terminator. So "hi" is actualy 'h' 'i' '\0'.

If you have a char array of just 'h', 'i' yes you will copy in whatever is after that until max is reached.

repository/gridftp/globus5/authz/source/Config.c
```diff
@@ -0,0 +1,497 @@
...
+ */
+bool validConfigLine(const char *line, int line_number, char *key,
+                     char *value) {
+  // Trim leading whitespace
+  while (isspace(*line)) {
+    line++;
+  }
+
+  // Skip empty lines or comments
+  if (*line == '\0' || *line == '#') {
```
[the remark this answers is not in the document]
JoshuaSBrown: This is partly why I have the C++ code mixed in as soon as possible. I am not aware of standard string support in the C language. The string.h file only provides utility functions i.e. strlen.

repository/gridftp/globus5/authz/source/Config.c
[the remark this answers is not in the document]
t-ramz: I had found incorrect documentation on `strncpy`, [GeeksForGeeks](https://www.geeksforgeeks.org/strncpy-function-in-c/) had said the final parameter was the number of characters copied, not the maximum characters copied.

[C++ Ref](https://cplusplus.com/reference/cstring/strncpy/), on the other hand says it's max characters.

Which leads me to another question:
Where are some good online docs for the C standard library?

repository/gridftp/globus5/authz/source/Config.c
t-ramz: Yeah this is a great catch, I didn't think about the possibility of running out of space for the null terminator during copy.
I'm not sure if it's more clear to write `len + 1 > a_max_len` or `len >= a_max_len` or `size_t len = strlen(a_src) + 1;` but whichever way I feel like it should be noted somewhere nearby the reason for a change.

repository/gridftp/globus5/authz/source/AuthzWorker.cpp
```diff
@@ -585,12 +589,20 @@ int AuthzWorker::checkAuth(char *client_id, char *path, char *action) {
...
+ 
+  int attempt = 0; 
+  int retries = 3;
+  do { 
+    DL_INFO(m_log_context,
+            "Sending RepoAuthzRequest client_id: " << client_id << " path: " << path << " action: " << action << " attempt: " << attempt);
+    m_comm->send(*message);
+    response = m_comm->receive(MessageType::GOOGLE_PROTOCOL_BUFFER);
+    ++attempt;
+  } while (response.time_out && attempt < retries);
```
t-ramz, line 605: I like having retries. Is there something that prompted this?

repository/gridftp/globus5/authz/source/AuthzWorker.cpp
```diff
@@ -629,26 +629,46 @@ const char *getReleaseVersion() {
...
+  if (log_path_authz2.length() > 0) {
+    // Append to the existing path because we don't want the C++ and C code
+    // trying to write to the same file
+    log_file_worker2.open(log_path_authz2, std::ios::app);
+    if (!log_file_worker2.is_open()) {
+      DL_ERROR(log_context, "AuthzWorker open log file path failed, path: " << log_path_authz2);
+    }
+
+    SDMS::global_logger.addStream(log_file_worker2);
+  }
```
t-ramz, line 668: Adding a comment so this Debug code doesn't get left in PR, or gets better naming.

repository/gridftp/globus5/authz/source/Config.c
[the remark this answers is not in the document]
JoshuaSBrown: C++ ref is the best general reference place. :)

repository/gridftp/globus5/authz/source/Config.c
JoshuaSBrown: This is a good point! Thanks for pointing it out. Since you brought up having single returns I realize the benefit more and more in terms of reducing errors and complexity.

repository/gridftp/globus5/authz/source/AuthzWorker.cpp
JoshuaSBrown: Ya, I was noticing messages weren't being passed to the core service and thought it might be a network issue. This should probably be customizable, but it's out of scope for this PR. The problem with having to many retries is it might just cause the GridFTP service to hang for a long time.

repository/gridftp/globus5/authz/source/AuthzWorker.cpp
JoshuaSBrown: What debug code?

JoshuaSBrown:
This PR is going to be split up into smaller pieces and merged more piecemeal as it has expanded beyond the initial scope.

JoshuaSBrown:
https://github.com/ORNL/DataFed/pull/1274

PR #1284, opened 2025-02-04
Feature DAPS 1215 foxx UserGetAccessTokenRequest mapped collection support

### LINKED ISSUE #1215: [DAPS, Foxx, Core] - Feature, adjust UserGetAccessRequest to work with Mapped Collection UUID
# Description

This is in support of moving document to and from mapped collections. We need someway of checking to make sure that a token exists if a mapped collection is being used.

Does a token already exist if it is a mapped collection or do we need to trigger a consent flow.
To avoid latency issues, we should only check if the token is expired and not make calls to Globus.

Below is relevant content in the datafed-ws.js file, this should be adjusted in a separte issue, it is shown here to indicate the connection between the web server and the core service.
```
app.get('/ui/ep/dir/list', ( a_req, a_resp ) => {

    // Aaron should do some work here to determine the collection type

    // if it is a mapped collection – might need if statement, need to check if we need to go through consent flow or not
    // if it is a mapped collection Collection UUID and Collection type need to be included in UserGetAccessTokenRequest message
    // If token is already stored and not expired needs to be returned here.
    // If token is expired need to indicate if token is valid
    sendMessage( "UserGetAccessTokenRequest", {}, a_req, a_resp, function( reply ) {

        const opts = {
            hostname: 'transfer.api.globusonline.org',
            method: 'GET',
            path: '/v0.10/operation/endpoint/' + encodeURIComponent(a_req.query.ep) + '/ls?path=' + encodeURIComponent(a_req.query.path) + '&show_hidden=' + a_req.query.hidden,
            rejectUnauthorized: true,
            headers:{
                Authorization: ' Bearer ' + reply.access
            }
        };

        const req = https.request( opts, (res) => {
            var data = '';

            res.on('data', (chunk) => {
                data += chunk;
            });
            res.on('end', () => {
                a_resp.json(JSON.parse(data));
            });
        });

        req.on('error', (e) => {
            a_resp.status( 500 );
            a_resp.send( "Globus endpoint directoy listing failed." );
        });

        req.end();
    });

});
```
Four branches – 

If token is in database for mapped collection, check expiration if valid return token, needs_consent bool is false
If token is in database for mapped collection, check expiration, if invalid use refresh token to create new access token, save in database, return needs_consent bool false
If token is in database for mapped collection, check expiration, if invalid refresh token fails, return needs_consent true
If token is not in database, return needs_consent: true

What needs to go in this if the access token does not exist for a mapped collection

message UserAccessTokenReply
{
    required string             access      = 1; // Globus access token
    required uint32             expires_in  = 2; // Access token expiration in seconds
}

# Acceptance

The four branches described above should work and have valid test cases.

t-ramz:
Closes #1215 

# PR Description
Add support for mapped collections when using `UserGetAccessTokenRequest` message to fetch user tokens.
Additionally, stub support for HA collections via comments and query parameters.

# Tasks

* [x] - CHANGELOG comment has been added
* [x] - If new feature a unit test has been added

## Summary by Sourcery

Add support for mapped collections to the UserGetAccessTokenRequest message.

New Features:
- Added support for mapped collections when fetching user tokens using the UserGetAccessTokenRequest message.

Tests:
- Added unit tests for the new mapped collection support in the user_router Foxx microservice.


core/server/DatabaseAPI.hpp
```diff
@@ -62,7 +62,12 @@ class DatabaseAPI {
                           const std::string &a_refresh_token,
                           LogContext log_context);
   void userGetAccessToken(std::string &a_acc_tok, std::string &a_ref_tok,
-                          uint32_t &a_expires_in, LogContext log_context);
+                          uint32_t &a_expires_in,
+                          const std::string collection_id,
+                          const std::string collection_type,
+                          bool &needs_consent,
+                          int &token_type, // TODO: use underlying type?
```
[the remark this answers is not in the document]
t-ramz: I had trouble converting from the numeric type I get as a return from JSON deserialization to the `AccessTokenType` that is represented here as an `int`. I chose to go the path of simply representing this as an `int` for now, which is the direct type that `AccessTokenType` (which is unfortunately not an enum class) compiles to in order to make the single comparison I absolutely need.

core/database/CMakeLists.txt
```diff
@@ -34,4 +35,6 @@ if( ENABLE_FOXX_TESTS )
   set_tests_properties(foxx_repo PROPERTIES FIXTURES_REQUIRED Foxx)
   set_tests_properties(foxx_path PROPERTIES FIXTURES_REQUIRED Foxx)
   set_tests_properties(foxx_user_router PROPERTIES FIXTURES_REQUIRED "Foxx;FoxxDBFixtures")
+  # may not require fixture
+  set_tests_properties(foxx_unit_user_token PROPERTIES FIXTURES_REQUIRED Foxx)
```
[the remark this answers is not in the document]
t-ramz: I'm unsure whether the Foxx services also provide the chai testing, but I believe they do. I'll look into it.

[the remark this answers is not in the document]
t-ramz: It's necessary, removing comment.

core/database/foxx/api/user_router.js
```diff
@@ -713,17 +718,40 @@ router
...
-                console.log("tok/get", Math.floor(Date.now() / 1000), user.expiration, exp);
-                result.expires_in = exp > 0 ? exp : 0;
-            } else {
-                console.log("tok/get - no expiration");
-                result.expires_in = 0;
+            let token_document = user;
+            let needs_consent = false;
+            if (collection_token) {
+                const globus_collection = g_db.globus_coll.exists({ _key: collection_id });
+                // TODO: should this be a query?
```
t-ramz, line 725: Another alternative is to use
```js
const token_matches = g_db.globus_token.byExample({
    _from: user._id
    _to: globus_collection?._id,
}).toArray();
```

The change would require a pre-check of `globus_collection`'s existence, otherwise Arango will throw an error.

But this may be both more readable and more performant.

t-ramz, line 725: Another possibility is the query:

```js
const token_matches = g_db._query(
    "FOR token_match " +
    "IN globus_token " +
    "FILTER token_match._from == @user_id " +
    "   AND token_match._to == @collection_id " +
    "RETURN token_match",
    {
        user_id: user._id,
        collection_id: globus_collection._id,
    }
).toArray();
```
This would require no other changes to logic. I'm not sure how much more performant or readable it is, however.

core/server/ClientWorker.cpp
```diff
@@ -1362,19 +1362,49 @@ std::unique_ptr<IMessage> ClientWorker::procUserGetAccessTokenRequest(
...
-
-    m_globus_api.refreshAccessToken(ref_tok, acc_tok, expires_in);
-    m_db_client.userSetAccessToken(acc_tok, expires_in, ref_tok, log_context);
+    if (token_type == AccessTokenType::GLOBUS_DEFAULT) {
+      m_globus_api.refreshAccessToken(ref_tok, acc_tok, expires_in);
+      m_db_client.userSetAccessToken(acc_tok, expires_in, ref_tok, log_context);
+    } else {
+      try {
+        m_globus_api.refreshAccessToken(ref_tok, acc_tok, expires_in);
+        m_db_client.userSetAccessToken(
```
t-ramz, line 1394: This attempts to cover condition 2, "If token is in database for mapped collection, check expiration, if invalid use refresh token to create new access token, save in database, return needs_consent bool false"

core/server/ClientWorker.cpp
```diff
@@ -1362,19 +1362,49 @@ std::unique_ptr<IMessage> ClientWorker::procUserGetAccessTokenRequest(
...
   m_db_client.setClient(a_uid);
-  m_db_client.userGetAccessToken(acc_tok, ref_tok, expires_in, log_context);
+  m_db_client.userGetAccessToken(acc_tok, ref_tok, expires_in, collection_id,
+                                 collection_type, needs_consent, token_type,
+                                 scopes, log_context);
 
-  if (expires_in < 300) {
+  if (needs_consent) {
+    // short circuit to reply
+  } else if (expires_in < 300) {
```
t-ramz, line 1386: This attempts to cover condition 1, "If token is in database for mapped collection, check expiration if valid return token, needs_consent bool is false"

core/database/foxx/api/user_router.js
```diff
@@ -713,17 +718,40 @@ router
...
+                                globus_collection?._id,
+                        ];
+                    }
+                    // TODO: account for AccessTokenType; currently only GLOBUS_DEFAULT and GLOBUS_TRANSFER are supported
+                    token_document = token_matches[0];
+                    needs_consent = false;
+                } else {
+                    // Cases covered: no matches, no tokens, no collection - all require same consent flow
+                    token_document = {};
+                    needs_consent = true;
```
t-ramz, line 747: This attempts to cover condition 4, "If token is not in database, return needs_consent: true"

core/database/foxx/api/user_router.js
```diff
@@ -713,17 +718,40 @@ router
...
+                            g_lib.ERR_INTERNAL_FAULT,
+                            "Too many matching tokens for user: " +
+                                user._id +
+                                " to collection: " +
+                                globus_collection?._id,
+                        ];
+                    }
+                    // TODO: account for AccessTokenType; currently only GLOBUS_DEFAULT and GLOBUS_TRANSFER are supported
+                    token_document = token_matches[0];
+                    needs_consent = false;
```
t-ramz, line 743: This attempts to cover condition 1, "If token is in database for mapped collection, check expiration if valid return token, needs_consent bool is false"

core/database/foxx/api/user_router.js
AronPerez: I’m a fan of using built-in methods to accomplish tasks. Raw queries nowadays are only necessary if

t-ramz: Agreed. Do you think that the `byExample` method or `outEdges` method and then `filter` on the resulting array is more readable?

core/database/foxx/api/lib/user_token.js
```diff
@@ -0,0 +1,89 @@
...
+                    "/token/get Requires 'collection_id' and 'collection_type' both if one is present, received " +
+                        "collection_id: " +
+                        collection_id +
+                        " collection_type: " +
+                        collection_type,
+                ];
+            }
+        }
+        return collection_token;
+    }
```
AronPerez, line 34: Really like this 🚀 

We can save some sanity and probably standardize validation by using something like [valibot](https://valibot.dev/guides/use-cases/).

I've mainly used [joi](https://github.com/hapijs/joi) but that's slowly spinning down. 

It's got a small footprint and we could build out the requirements for our classes in them and validate them in middleware.

Check out https://valibot.dev/guides/quick-start/

AronPerez: Another good one right now seems to be [zod](https://zod.dev/)

core/database/foxx/api/lib/user_token.js
```diff
@@ -0,0 +1,89 @@
...
+            scopes: "",
+        };
+        if (needs_consent) {
+            // short circuit when consent flow is required
+            return result_token;
+        }
+        result_token.access = token_document.access;
+        result_token.refresh = token_document.refresh;
+        if (token_document.expiration) {
+            const expires_in = token_document.expiration - Math.floor(Date.now() / 1000);
```
AronPerez, line 73: What does Math.floor(Date.now() / 1000) corrospond to

common/proto/common/SDMS_Auth.proto
```diff
@@ -124,14 +124,17 @@ message UserSetAccessTokenRequest
 // Reply: UserAccessTokenReply on success, NackReply on error
 message UserGetAccessTokenRequest
 {
+    optional string             collection_id   = 1;
+    optional string             collection_type = 2;    // TODO: use enum
```
JoshuaSBrown, line 128: Yes

JoshuaSBrown: Did we not discuss the enums somewhere already?

core/database/foxx/api/user_router.js
AronPerez: I like `byExample`. Seems straightforward for what we need

```javascript
const user = g_db.u.byExample({ _id: req.queryParams.subject }).toArray();  
``

File: core/database/foxx/api/user_router.js [hunk 11]
```diff
@@ -732,6 +760,16 @@ router
     })
     .queryParam("client", joi.string().required(), "Client ID")
     .queryParam("subject", joi.string().optional(), "UID of subject user")
+    .queryParam(
+        "collection_id",
+        joi.string().optional().guid(),
+        "ID of collection with which token is associated",
+    ) // https://joi.dev/api/?v=17.13.3#stringguid---aliases-uuid
```
AronPerez, line 767: Oh I didn't realize we were using joi here, ignore my comment on using zod/valibot

common/proto/common/SDMS_Auth.proto
JoshuaSBrown: Not in this pr though.

core/database/foxx/api/lib/user_token.js
AronPerez: https://github.com/ORNL/DataFed/pull/1284#discussion_r1945943383

common/proto/common/SDMS_Auth.proto
```diff
@@ -124,14 +124,17 @@ message UserSetAccessTokenRequest
...
 
 // Reply containing the user Globus access token and expiration time (in
 // seconds) for authenticated user.
 message UserAccessTokenReply
 {
-    required string             access      = 1; // Globus access token
-    required uint32             expires_in  = 2; // Access token expiration in seconds
+    required string             access          = 1; // Globus access token
+    required uint32             expires_in      = 2; // Access token expiration in seconds
+    required bool               needs_consent   = 3; // Indicate requirement of consent flow
```
JoshuaSBrown, line 137: If you make this required you will break the API, can we get away with keeping this optional?

When I say break the API, I mean that it won't be backwards compatible with how it is implemented currently.

JoshuaSBrown: In proto3 they have actually moved to using optional by default and avoided required for this reason.

core/database/CMakeLists.txt
```diff
@@ -22,6 +22,7 @@ if( ENABLE_FOXX_TESTS )
   add_test(NAME foxx_support COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/tests/test_foxx.sh" -t "unit_support")
   add_test(NAME foxx_user_router COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/tests/test_foxx.sh" -t "unit_user_router")
   add_test(NAME foxx_authz_router COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/tests/test_foxx.sh" -t "unit_authz_router")
+  add_test(NAME foxx_unit_user_token COMMAND "${CMAKE_CURRENT_SOURCE_DIR}/tests/test_foxx.sh" -t "unit_user_token")
```
JoshuaSBrown, line 25: Nice!

core/database/foxx/api/lib/user_token.js
t-ramz: This was a copy/paste from https://github.com/ORNL/DataFed/blob/devel/core/database/foxx/api/user_router.js#L720

This is normalizing the data stored on the database from the expiry date/time to a time until expiry.

core/database/foxx/api/user_router.js
t-ramz: I also like that, and I found other uses of `byExample` in the Foxx API, so I think it fits in better as well. Will make that change 🫡

common/proto/common/SDMS_Auth.proto
t-ramz: Good point. With that in mind, all code that can point to a mapped collection will need to validate both the field's existence and the field's value

core/database/foxx/api/user_router.js
t-ramz: This has been implemented, along with an additional test to look for the missing collection case!

common/proto/common/SDMS_Auth.proto
t-ramz: I made it optional, do you have an idea on how we can enforce some diligence on checks?

core/database/foxx/api/user_router.js
t-ramz: Yeah, that was the old behavior. Brought that back.

common/proto/common/SDMS_Auth.proto
JoshuaSBrown: When the message comes into the core service, whether from the web server or the python client, it will go through the DatabaseAPI class. That would be a good place.

AronPerez (approved):

Issue #1285, opened 2025-02-04
[Bug] - DataFed Schema for empty metadata field


JoshuaSBrown:
# Description

A schema applied to a metadata record does not trigger warnings if no text is placed in metadata tab.

![Image](https://github.com/user-attachments/assets/6244bf93-3dd4-4d8e-a580-681bf0482eb7)

![Image](https://github.com/user-attachments/assets/8b3186a6-3b15-47b6-a5a8-a3b8e3ea5b4f)

Notice below, though schema is applied no warnings are given if no metadata has been added empty {} would trigger the warnings.

![Image](https://github.com/user-attachments/assets/07f509f8-f3c4-4271-8e27-555114854352)

# Acceptance

Fix create metadata by adding empty curly braces if on empty so warnings are triggered.

PR #1328, opened 2025-02-20
JSON parse error fix

t-ramz:
# PR Description
Security fixes

# Tasks

## Summary by Sourcery

This pull request replaces the custom JSON serialization with the nlohmann/json library to address potential security vulnerabilities and improve code maintainability. The changes affect several database API methods, including record creation, updates, and searches.

Bug Fixes:
- Fixes a bug where JSON parsing was not handled correctly in several database API methods.

Enhancements:
- Replaces the custom JSON serialization with the nlohmann/json library for improved reliability and security.


core/server/DatabaseAPI.cpp
```diff
@@ -1737,12 +1734,22 @@ void DatabaseAPI::queryCreate(const Auth::QueryCreateRequest &a_request,
     EXCEPT(1, "Invalid search request");
   }
 
-  string body =
+  /*string body =
```
t-ramz, line 1737: oops this needs to be removed

core/server/DatabaseAPI.cpp
```diff
@@ -1278,13 +1268,24 @@ void DatabaseAPI::generalSearch(const Auth::SearchRequest &a_request,
...
                                     params, log_context);
 
-  string body =
-      "{\"mode\":" + to_string(a_request.mode()) + ",\"published\":" +
-      ((a_request.has_published() && a_request.published()) ? "true"
-                                                            : "false") +
-      ",\"qry_begin\":\"" + qry_begin + "\",\"qry_end\":\"" + qry_end +
-      "\",\"qry_filter\":\"" + qry_filter + "\",\"params\":{" + params +
-      "},\"limit\":" + to_string(cnt) + "}";
+  /* string body =
```
t-ramz, line 1271: this also needs removed

core/server/DatabaseAPI.cpp
```diff
@@ -1776,13 +1786,22 @@ void DatabaseAPI::queryUpdate(const Auth::QueryUpdateRequest &a_request,
       EXCEPT(1, "Invalid search request");
     }
 
-    body += ",\"qry_begin\":\"" + qry_begin + "\",\"qry_end\":\"" + qry_end +
+    /*body += ",\"qry_begin\":\"" + qry_begin + "\",\"qry_end\":\"" + qry_end
```
t-ramz, line 1789: also needs removal

core/server/DatabaseAPI.cpp
```diff
@@ -3111,6 +3086,7 @@ void DatabaseAPI::taskAbort(const std::string &a_task_id,
                             libjson::Value &a_task_reply,
                             LogContext log_context) {
   libjson::Value doc(a_msg);
+  // TODO: json serialization --FLAG CHECK ON THIS LATER
```
t-ramz, line 3089: this will require some further investigation that I need access to on-site network to facilitate. I don't know, sort of at all, what `a_msg` looks like

core/server/DatabaseAPI.cpp
```diff
@@ -1043,16 +1031,19 @@ void DatabaseAPI::recordUpdateSize(const Auth::RepoDataSizeReply &a_size_rep,
...
-    if (i > 0)
-      body += ",";
-    body += "{\"id\":\"" + a_size_rep.size(i).id() +
-            "\",\"size\":" + to_string(a_size_rep.size(i).size()) + "}";
+  nlohmann::json records = nlohmann::json::array();
+  for (int i = 0; i < a_size_rep.size_size(); i++) { // size size :D
+    nlohmann::json record_entry;
+    record_entry["id"] = a_size_rep.size(i).id();
+    record_entry["size"] =
+        to_string(a_size_rep.size(i).size()); // size size size!! :D
```
t-ramz, line 1041: need to `push_back` to `records`

core/server/DatabaseAPI.cpp
```diff
@@ -1737,12 +1726,16 @@ void DatabaseAPI::queryCreate(const Auth::QueryCreateRequest &a_request,
...
-      params + "},\"limit\":" + to_string(cnt) + ",\"title\":\"" +
-      escapeJSON(a_request.title()) + "\"" + ",\"query\":" + query_json + "}";
+  nlohmann::json payload;
+  payload["qry_begin"] = qry_begin;
+  payload["qry_end"] = qry_end;
+  payload["qry_filter"] = qry_filter;
+  payload["params"] = params;
+  payload["limit"] = to_string(cnt);
+  payload["title"] = escapeJSON(a_request.title());
+  payload["query"] = query_json;
```
t-ramz, line 1736: I'm not sure if this could break when getting "double" serialized

core/server/DatabaseAPI.cpp
```diff
@@ -2863,109 +2854,70 @@ void DatabaseAPI::schemaView(const Auth::SchemaViewRequest &a_request,
...
-    body.append(a_request.pub() ? "true" : "false");
+    payload["pub"] = a_request.pub();
   }
 
   if (a_request.has_sys()) {
-    if (body.size() > 1)
-      body.append(",");
-
-    body.append("\"sys\":");
-    body.append(a_request.sys() ? "true" : "false");
```
t-ramz, line 2905: May be worth investigating if `.sys()` is a boolean or not, because if this is checking truthiness and we're passing the actual value now, the backend could behave differently.

core/server/DatabaseAPI.cpp
```diff
@@ -3121,29 +3074,27 @@ void DatabaseAPI::taskInitDataGet(const Auth::DataGetRequest &a_request,
...
-    body += ",\"path\":\"" + a_request.path() + "\"";
+    payload["path"] = a_request.path();
 
   if (a_request.has_encrypt())
-    body += ",\"encrypt\":" + to_string(a_request.encrypt());
+    payload["encrypt"] = to_string(a_request.encrypt());
 
   if (a_request.has_orig_fname() && a_request.orig_fname())
-    body += ",\"orig_fname\":true";
+    payload["orig_fname"] = a_request.orig_fname();
```
t-ramz, line 3092: depends on type of `orig_fname()`

core/server/DatabaseAPI.cpp
```diff
@@ -3245,16 +3198,9 @@ void DatabaseAPI::setDataPutReply(Auth::DataPutReply &a_reply,
...
-  for (vector<string>::const_iterator i = a_ids.begin(); i != a_ids.end();
-       ++i) {
-    if (i != a_ids.begin())
-      body += ",";
-
-    body += "\"" + *i + "\"";
-  }
-  body += "]}";
+  nlohmann::json payload;
+  payload["ids"] = a_ids;
```
t-ramz, line 3202: Direct assignment should be possible here because `vec<std::string>` is a type `nlohmann/json` implicitly converts. https://json.nlohmann.me/features/arbitrary_types/

core/server/DatabaseAPI.cpp
t-ramz: On further investigation, it looks like `taskAbort` is never called or assigned, only declared. I believe this is dead code.

core/server/DatabaseAPI.cpp
```diff
@@ -885,64 +886,53 @@ void DatabaseAPI::recordCreate(const Auth::RecordCreateRequest &a_request,
...
-  body.append(escapeJSON(a_request.title()));
-  body.append("\"");
+  payload["title"] = escapeJSON(a_request.title());
 
   if (a_request.has_desc())
-    body += ",\"desc\":\"" + escapeJSON(a_request.desc()) + "\"";
+    payload["desc"] = escapeJSON(a_request.desc());
   if (a_request.has_alias())
-    body += ",\"alias\":\"" + a_request.alias() + "\"";
+    payload["alias"] = a_request.alias();
```
JoshuaSBrown, line 896: Doesn't this need to be escaped as well? Are you just trying to make as few changes as possible? If so this needs to probably be made into an issue and investigated as an attack vector as well.

core/server/DatabaseAPI.cpp
```diff
@@ -885,64 +886,53 @@ void DatabaseAPI::recordCreate(const Auth::RecordCreateRequest &a_request,
...
+    nlohmann::json deps = nlohmann::json::array();
 
     for (int i = 0; i < a_request.deps_size(); i++) {
-      body += string(i > 0 ? "," : "") + "{\"id\":\"" + a_request.deps(i).id() +
-              "\",\"type\":" + to_string(a_request.deps(i).type()) + "}";
+      nlohmann::json dep_entry;
+      dep_entry["id"] = a_request.deps(i).id();
+      dep_entry["type"] = to_string(a_request.deps(i).type());
+      deps.push_back(dep_entry);
     }
```
JoshuaSBrown, line 932: This can be converted to using a range based for loop and emplace back to avoid a copy.

```
for (const auto& dep : a_request.deps()) {
    deps.emplace_back(nlohmann::json{
        {"id", dep.id()},
        {"type", to_string(dep.type())}
    });
}
```

core/server/DatabaseAPI.cpp
```diff
@@ -964,65 +954,63 @@ void DatabaseAPI::recordCreateBatch(
...
-    body += ",\"tags_clear\":true";
+    payload["tags_clear"] = a_request.tags_clear();
   } else if (a_request.tags_size()) {
-    body += ",\"tags\":[";
+    nlohmann::json tags = nlohmann::json::array();
     for (int i = 0; i < a_request.tags_size(); i++) {
-      if (i)
-        body += ",";
-      body += "\"" + a_request.tags(i) + "\"";
+      tags.push_back(a_request.tags(i));
```
JoshuaSBrown, line 971: I prefer always wrapping if statement with {} to prevent accidental newlines breaking things.

JoshuaSBrown: Would again recommed using range based for

```
for (const auto& tag : a_request.tags()) {
    tags.push_back(tag);
}
```

core/server/DatabaseAPI.cpp
```diff
@@ -964,65 +954,63 @@ void DatabaseAPI::recordCreateBatch(
...
     }
-    body += "]";
+    payload["tags"] = tags;
   }
 
   if (a_request.has_metadata()) {
-    body += ",\"md\":" +
-            (a_request.metadata().size() ? a_request.metadata() : "\"\"");
+    payload["md"] = a_request.metadata();
```
JoshuaSBrown, line 977: Is this the same? What does metadata() return if it is empty?

core/server/DatabaseAPI.cpp
```diff
@@ -3185,21 +3137,23 @@ void DatabaseAPI::taskInitDataPut(const Auth::DataPutRequest &a_request,
...
   if (a_request.has_ext())
-    body += ",\"ext\":\"" + a_request.ext() + "\"";
+    payload["ext"] = a_request.ext();
 
   if (a_request.has_check() && a_request.check())
-    body += ",\"check\":true";
+    payload["check"] = a_request.check();
 
-  body += "}";
+  string body = payload.dump();
```
AronPerez, line 3156: So we escape chars with `.dump()` by default.

Any reason to not escape Non-ASCII chars with [ensure_ascii](https://json.nlohmann.me/api/basic_json/dump/#parameters) param?

core/server/DatabaseAPI.cpp
JoshuaSBrown: I think this actually might be relaged to this schema bug. It looks like nothing is passed back if has_metdata is false: https://github.com/ORNL/DataFed/issues/1285

AronPerez (commented):
1 question

core/server/DatabaseAPI.cpp
t-ramz: No reason, I can gladly add that if we think it's helpful. Would likely help to further mitigate the attack vectors.

core/server/DatabaseAPI.cpp
t-ramz: Yup, few changes as possible. `nlohmann:/json` will escape any JSON control values we don't explicitly escape, so any of the `escapeJSON` calls are likely redundant.

core/server/DatabaseAPI.cpp
t-ramz: Can I construct the map type in C++ in place like that?

core/server/DatabaseAPI.cpp
t-ramz: Should be an empty string as initialized by protobuf

t-ramz:
Got a few questions about validating, so here's what I did, and is too much lift to reproduce in a unit test for a patch:

1. Follow this guide to set up a quick c++ project: https://docs.docker.com/guides/cpp/develop/
2. Download the latest header from https://github.com/nlohmann/json/releases
3. Copy file into image by editing the Dockerfile
4. Import into code
5. Add on get some malicious JSON (just in code)
6. Write out the malicious json.dump() to response
7. Hit endpoint to verify it is escaped

core/server/DatabaseAPI.cpp
JoshuaSBrown: Looks like you have to have the nlohmann::json object. Which was missing before I updated the comment.

core/server/DatabaseAPI.cpp
JoshuaSBrown: I see thanks for clarifying.

core/server/DatabaseAPI.cpp
JoshuaSBrown: https://github.com/ORNL/DataFed/blob/1cf4921eedb7bc707467e5e4010b2a0d1d422edd/common/proto/common/SDMS_Auth.proto#L1327

I think you already verified this, but you did it right.

core/server/DatabaseAPI.cpp
```diff
@@ -2863,109 +2855,70 @@ void DatabaseAPI::schemaView(const Auth::SchemaViewRequest &a_request,
...
-    if (body.size() > 1)
-      body.append(",");
-
-    body.append("\"def\":");
-    body.append(a_request.def());
+    payload["def"] = a_request.def();
   }
-
-  body.append("}");
+  string body = payload.dump();
```
JoshuaSBrown, line 2921: Huge improvement in quality and readability.

core/server/DatabaseAPI.cpp
JoshuaSBrown: Boolean https://github.com/ORNL/DataFed/blob/1cf4921eedb7bc707467e5e4010b2a0d1d422edd/common/proto/common/SDMS_Auth.proto#L431

JoshuaSBrown (approved):
Overall, a huge improvement in code quailty that improves the security posture of the code base.

Possible optional improvements include using range based for loops to reduce verbosity, and the addition of curly braces around if statements. 

Approved on passing the CI.

AronPerez (approved):

JoshuaSBrown (commented):
Wait let me look at this function.

google::protobuf::util::MessageToJsonString

core/server/DatabaseAPI.cpp
JoshuaSBrown: Yes, we have to remove all escapeJSON calls that are being used before this can be merged, it leads to double escapes.

```
Before using nlohman
{"query":{
 "id": 42,
 "name": "Alice",
 "is_active": true
}
}

After using nlohmann
{"query":"{\n \"id\": 42,\n \"name\": \"Alice\",\n \"is_active\": true\n}\n"}

Using nlohmann with parse
{"query":{"id":42,"is_active":true,"name":"Alice"}}

Title without nlohmann json but with escapeJSON
{"title": "This is my \"Great Title\"" }

title with nlohmann json and escapeJSON call
{"title":"This is my \\\"Great Title\\\""}

title with nlohmann json but no escapeJSON call
{"title":"This is my \"Great Title\""}
```

JoshuaSBrown (changes requested):
Must remove all escapeJSON calls that are being passed into nlohmann json object.

JoshuaSBrown:
Files to reproduce testing. 
[test_proto_json.zip](https://github.com/user-attachments/files/18943023/test_proto_json.zip)

core/server/DatabaseAPI.cpp
```diff
@@ -1737,12 +1752,16 @@ void DatabaseAPI::queryCreate(const Auth::QueryCreateRequest &a_request,
...
-      params + "},\"limit\":" + to_string(cnt) + ",\"title\":\"" +
-      escapeJSON(a_request.title()) + "\"" + ",\"query\":" + query_json + "}";
+  nlohmann::json payload;
+  payload["qry_begin"] = qry_begin;
+  payload["qry_end"] = qry_end;
+  payload["qry_filter"] = qry_filter;
+  payload["params"] = params;
+  payload["limit"] = to_string(cnt);
+  payload["title"] = a_request.title();
+  payload["query"] = nlohmann::json::parse(query_json);
```
t-ramz, line 1762: I'm not totally sure what the original protobuf message looked like, but does it make sense to serialize to JSON, deserialize, then serialize again?

JoshuaSBrown: Because of things like this, I am fine with the back and forth between serializing and deserializing, I think it is safer to rely on protobufs built in functions then to code this our selves. 

```
options.always_print_enums_as_ints = true;
options.preserve_proto_field_names = true;
```

core/server/DatabaseAPI.cpp
```diff
@@ -3659,6 +3705,29 @@ void DatabaseAPI::metricsUpdateMsgCounts(
...
+  string body;
+  string new_body = newJsonMetricParse(a_timestamp, a_total, a_metrics);
+  string old_body = oldJsonMetricParse(a_timestamp, a_total, a_metrics);
+
+  if (new_body == old_body) {
+    // on match use safer serialization
+    body = new_body;
+  } else {
+    body = old_body;
+    DL_DEBUG(
```
t-ramz, line 3725: Should we log a warning or error instead so that it will be very visible on deployments that there is a potential risk?

JoshuaSBrown: Yes, I would log a warning.

JoshuaSBrown: It's not clear if it would be an ERROR if it is just a white space difference.

core/server/DatabaseAPI.cpp
t-ramz: This will be included, and I want to leave the comment that the only reason we may run into issues is if we have UTF encoded strings at any point that we would like to support by default. I am not sure exactly what incoming data will look like necessarily.

JoshuaSBrown (approved):

PR #1350, opened 2025-03-26
[DAPS-1349] Provenance Capture Update Bug

AronPerez:
# PR Description

### Manually
1. Log in
2. Select a folder/collection
3. Create a record with base information, no relationships defined.
4. Submit the record for creation
5. Select the record and click edit

### Before
https://github.com/user-attachments/assets/8ad42f13-9349-4951-ba96-ab6964307e93

### After
https://github.com/user-attachments/assets/7eed4c5a-c5e7-4fdc-aead-d4affc7e906c

## Summary by Sourcery

Fix dependency comparison logic in record creation and editing process

Bug Fixes:
- Correct the logic for comparing dependencies to prevent incorrect handling of record relationships

Enhancements:
- Refactor dependency comparison checks to improve accuracy and prevent potential errors in record dependency management


JoshuaSBrown (commented):
Please would you add a comment around these if statements about how strict equality is intentionally not wanted here. I believe this was changed in the first place because static code analyzers suggest using strict equality. It would also be nice to know exactly why strict equality causes the bug.

AronPerez:
## Summary by Sourcery

Fix logic in dependency comparison to correctly identify and handle dependencies in data editing

Bug Fixes:
- Corrected the logic for comparing dependencies to accurately detect unique and duplicate dependencies in the data editing process

Enhancements:
- Refactored dependency comparison logic to use more explicit boolean return conditions

web/static/dlg_data_new_edit.js
```diff
@@ -410,7 +410,8 @@ export function show(a_mode, a_data, a_parent, a_upd_perms, a_cb) {
 
                             if (
                                 deps.findIndex(function (el) {
-                                    return !(dep.id !== el.id || dep.type !== el.type);
+                                    if (dep.id != el.id || dep.type != el.type) return false;
+                                    else return true;
```
t-ramz, line 414: Is this just to revert it?
I would prefer we keep the `return !(...)` or do what sourcery suggested, but I can also get behind seeing no diff on this file from ~3mo ago to now.

t-ramz (approved):
Small comment. Looks good.

AronPerez:
```javascript
idx === i  // FALSE
```

```javascript
idx == i  // TRUE
```

Screenshot 2025-03-27 at 3 16 19 PM

Screenshot 2025-03-27 at 3 18 53 PM

PR #1432, opened 2025-05-15
Austin token encryption

JoshuaSBrown:
# PR Description

# Tasks


common/source/CipherEngine.cpp
```diff
@@ -0,0 +1,227 @@
...
+    void CipherEngine::handleErrors(void)
+    {
+        ERR_print_errors_fp(stderr);
+        abort();
+    }
+    
+    CipherEngine::CipherString CipherEngine::createCipherString()
+    {
+        CipherString cs;
+        return cs;
```
JoshuaSBrown, line 30:
```suggestion
        return CipherString();
```

JoshuaSBrown: Actually this whole method is redundant, because CipherString is public you can just do this. 
```c++
CipherEngine::CipherString cs;
```

Where ever you need it.

common/source/CipherEngine.cpp
```diff
@@ -0,0 +1,227 @@
...
+    }
+    
+    std::unique_ptr<char[]> encode64(const unsigned char* input, int length)
+    {
+        const int pl = 4*((length+2)/3);
```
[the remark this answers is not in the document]
JoshuaSBrown: Also pl is not a good choice of name, it is better to be a bit more verbose when creating variable names.

common/source/CipherEngine.cpp
```diff
@@ -0,0 +1,227 @@
...
+        auto output = std::make_unique<char[]>(pl+1);
+        std::fill_n(output.get(), pl+1, 0);  // manually zero-initialize 
+        //char* output = reinterpret_cast<char *>(calloc(pl+1, 1)); //+1 for the terminating null that EVP_EncodeBlock adds on
+        const int ol = EVP_EncodeBlock(reinterpret_cast<unsigned char*>(output.get()), input, length);
+        if (pl != ol) { std::cerr << "Whoops, encode predicted " << pl << " but we got " << ol << "\n"; } 
+        return std::move(output);
+    } 
+ 
+    std::unique_ptr<unsigned char[]> decode64(const char *input, int length) {
```
JoshuaSBrown, line 44: Same critique as for the encode function.

common/source/CipherEngine.cpp
```diff
@@ -0,0 +1,227 @@
...
+#include <memory>
+#include <string.h>
+#include <string>
+#include <vector>
+#include <fstream>
+#include <iomanip>
+using namespace std;
+namespace SDMS{
+
+    void CipherEngine::handleErrors(void)
```
JoshuaSBrown, line 21: Can this be made a const method?

common/source/CipherEngine.cpp
```diff
@@ -0,0 +1,227 @@
...
+    CipherEngine::CipherString CipherEngine::encrypt(const string& msg)
+    {
+        unsigned char iv[16] = {};
+        generateIV(iv);
+        
+        CipherString result;
+
+        result = encrypt_algorithm(iv, msg);
+        
+        return result;
```
JoshuaSBrown, line 160:
```suggestion
        return encrypt_algorithm(iv, msg);
```

common/source/CipherEngine.cpp
```diff
@@ -0,0 +1,227 @@
...
+    std::string CipherEngine::decrypt(const CipherString& encoded_encrypted_string)
+    {
+    
+    EVP_CIPHER_CTX *ctx = nullptr;
+
+    //unsigned char* ciphertext;
+    int ciphertext_len = encoded_encrypted_string.encrypted_msg_len;
+    //unsigned char* iv;
+
+    int len;
```
JoshuaSBrown, line 172: Move this down to where it is being used. Try to keep variables as closed to where they are being used as possible.

common/source/CipherEngine.cpp
```diff
@@ -0,0 +1,227 @@
...
+    unsigned char plaintext[encoded_encrypted_string.encrypted_msg_len + EVP_MAX_BLOCK_LENGTH] = {};
+    int plaintext_len;
```
JoshuaSBrown, line 174: Same for this.

common/source/CipherEngine.cpp
```diff
@@ -0,0 +1,227 @@
...
+    }
+    plaintext_len += len;
+
+    /* Clean up */
+    EVP_CIPHER_CTX_free(ctx);
+
+    //Convert the plaintext back to string
+    std::string result(reinterpret_cast<char const*>(plaintext), plaintext_len);
+       
+    return result;
```
JoshuaSBrown, line 225:
```suggestion
    return std::string(reinterpret_cast<char const*>(plaintext), plaintext_len);
```

common/source/Util.cpp
```diff
@@ -34,6 +39,21 @@ std::string exec(const char *cmd) {
   return result;
 }
 
+int readFile(std::string fileName, int arraySize, unsigned char* array)
```
JoshuaSBrown, line 42: make fileName a const reference, you shouldn't be changing it. 

Please also check to see if variables in the code are snake case, I'm pretty sure they are.

JoshuaSBrown: array size should also be const.

common/source/Util.cpp
```diff
@@ -34,6 +39,21 @@ std::string exec(const char *cmd) {
...
+    unsigned char keyChar[arraySize] = {};
+    //Grabbing key
+    std::ifstream keyFile(fileName, std::ios::binary);
+
+    keyFile.read(reinterpret_cast<char*>(keyChar),arraySize);
+
+    for (int lv = 0; lv < arraySize; lv++)
+    {
+        array[lv] = keyChar[lv];
+    }
```
JoshuaSBrown, line 54: Check and see if you can replace this with:

```
#include <algorithm> // for std::copy

std::copy(keyChar, keyChar + arraySize, array);
```

common/source/Util.cpp
```diff
@@ -34,6 +39,21 @@ std::string exec(const char *cmd) {
...
+return 1;
```
JoshuaSBrown, line 55: If it is always returning 1 why not make it void, or you could see if you can capture a return value from the read command.

common/tests/unit/test_CipherEngine.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+BOOST_AUTO_TEST_SUITE(KeyEncryptionDecryptionTest)
+
+BOOST_AUTO_TEST_CASE(test_EncryptionDecryption)
+{
+
+    std::cout << "THIS IS A NEW FLAG" << std::endl;
+    unsigned char key[32];
+    //CipherEngine::generateEncryptionKey(key);
+    readFile("../../build/core/server/datafed-token-key.txt", 32, key);
```
JoshuaSBrown, line 28: magic number

common/tests/unit/test_CipherEngine.cpp
```diff
@@ -0,0 +1,148 @@
...
+using namespace SDMS;
+
+BOOST_AUTO_TEST_SUITE(KeyEncryptionDecryptionTest)
+
+BOOST_AUTO_TEST_CASE(test_EncryptionDecryption)
+{
+
+    std::cout << "THIS IS A NEW FLAG" << std::endl;
+    unsigned char key[32];
+    //CipherEngine::generateEncryptionKey(key);
```
JoshuaSBrown, line 27:
```suggestion
```

common/tests/unit/test_CipherEngine.cpp
JoshuaSBrown: Also, would it be better to create the file as part of the test? This has hard coded in the build folder depending on what cmake command a user runs it may not exist 

```
cmake -S. -B different_build_folder
```

Then the build folder won't be called build.

JoshuaSBrown (commented):
I'm not done but it gives you some things to work on.

core/database/foxx/api/tasks.js
```diff
@@ -596,6 +600,11 @@ var tasks_func = (function () {
                 token_type: extra_token_format.token_type,
                 scopes: extra_token_format.scopes,
                 collection_id: state.collection_info.collection_id,
+                acc_tok_iv: tokens.acc_tok_iv,
+                acc_tok_len: tokens.acc_tok_len,
+                ref_tok_iv: tokens.ref_tok_iv,
+                ref_tok_len: tokens.ref_tok_len,
+
```
t-ramz, line 607: See my comment in user_token.js

core/database/foxx/api/user_router.js
```diff
@@ -553,6 +552,7 @@ router
                     var user_id;
                     let user_doc;
 
+                    console.log("Recieved Access Token:", JSON.stringify(req.queryParams.access));
```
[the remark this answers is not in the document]
t-ramz: It is best to remove these logs

core/database/foxx/api/user_router.js
```diff
@@ -117,7 +117,6 @@ router
                         ct: time,
                         ut: time,
                     };
-
```
t-ramz, line 120: Whitespace

t-ramz (changes requested):
I'm going to split my review into two parts since it's a lot of work (we should have split this up into multiple pieces!).

Anyway, I have added quite a few comments and I want to add a big comment about whitespace here:

We should limit whitespaces changes when possible, so that diffs are both more helpful for development and less cumbersome during reviews.

One way to do this is to use the `git checkout --patch` command here: https://git-scm.com/docs/git-checkout/2.5.6#Documentation/git-checkout.txt---patch

Which you could use like `git checkout --patch origin/devel <filename>` and then follow this section here for some help when you're thrown into an interactive patch: https://git-scm.com/docs/git-add#_interactive_mode

common/include/common/CipherEngine.hpp
```diff
@@ -0,0 +1,49 @@
+#ifndef CIPHER_ENGINE_HPP
+#define CIPHER_ENGINE_HPP
+#pragma once
+
+//Local include
+#include <string>
+#include <memory>
+
+namespace SDMS{
+class CipherEngine
```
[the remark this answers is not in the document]
t-ramz: Could you point us to what you're looking for specifically? I found https://www.doxygen.nl/manual/docblocks.html - is that right?

common/include/common/CipherEngine.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+        struct CipherString
+        {
+            std::unique_ptr<char[]> encrypted_msg;
+            std::unique_ptr<char[]> iv;
+            int encrypted_msg_len;
+        };
+
+        CipherString createCipherString();
+        CipherString encrypt_algorithm(unsigned char *iv, const std::string& msg);
```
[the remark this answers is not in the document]
t-ramz: In general, CamelCase is fine for OO things like - Class names and method names (if you're feeling like you hate everything). snake_case is a must for arguments and variables, however.

core/server/TaskWorker.cpp
```diff
@@ -190,12 +193,83 @@ void TaskWorker::workerThread(LogContext log_context) {
   } // End of outer while loop
 }
 
+
+
+std::string
+TaskWorker::prepToken(const Value::Object &obj,std::string token, bool needs_update, LogContext log_context)
```
t-ramz, line 199: `token` should have a more descriptive name. since it's either `acc_tok` (access_token) or `ref_tok` it could just be a bool `is_access_token`, for example.

core/server/TaskWorker.cpp
```diff
@@ -190,12 +193,83 @@ void TaskWorker::workerThread(LogContext log_context) {
...
+        std::memcpy(encoded_obj.iv.get(), iv_str.c_str(), iv_str.size() + 1);     // copy including '\0' 
+        
+
+        //Decrypt it:
+        string decrypted_token = cipher.decrypt(encoded_obj);
+        
+        return decrypted_token;
+    }
+    else
+    {
```
t-ramz, line 233: Could make sense to extract this logic to the caller. I know that function is already huge, however. Hard to say.

core/server/TaskWorker.cpp
```diff
@@ -207,8 +281,9 @@ TaskWorker::cmdRawDataTransfer(TaskWorker &me, const Value &a_task_params,
   bool encrypted = true;
   GlobusAPI::EndpointInfo ep_info;
 
-  string acc_tok = obj.getString("acc_tok");
-  string ref_tok = obj.getString("ref_tok");
+  //string acc_tok = obj.getString("acc_tok");
+  //string ref_tok = obj.getString("ref_tok");
+
```
t-ramz, line 286: let's remove these comments

core/server/TaskWorker.cpp
```diff
@@ -222,7 +297,7 @@ TaskWorker::cmdRawDataTransfer(TaskWorker &me, const Value &a_task_params,
 
   DL_TRACE(log_context, ">>>> Token Expires in: " << expires_in);
 
-  if (expires_in < 3600) {
+  if ((expires_in < 3600) | needs_update/*Add an or its missing certain fields*/) {
```
t-ramz, line 300: Why are we doing a bitwise `or` 😨

core/server/TaskWorker.cpp
```diff
@@ -427,6 +502,8 @@ ICommunicator::Response TaskWorker::cmdAllocCreate(TaskWorker &me,
                                                    LogContext log_context) {
   const Value::Object &obj = a_task_params.asObject();
 
+  DL_DEBUG(log_context, "Testing to see what thisll output");
```
t-ramz, line 505: Let's remove this

core/server/TaskWorker.hpp
```diff
@@ -23,12 +23,15 @@ namespace SDMS {
...
+protected:
+    static bool tokenNeedsUpdate(const libjson::Value::Object &obj);
+    static std::string prepToken(const libjson::Value::Object &obj, std::string token, bool needs_update, LogContext log_context);
 public:
   TaskWorker(ITaskMgr &a_mgr, uint32_t a_id, LogContext log_context);
   ~TaskWorker();
 
 private:
-  typedef ICommunicator::Response (*task_function_t)(
+    typedef ICommunicator::Response (*task_function_t)(
```
t-ramz, line 34: Whitespace

core/server/main.cpp
```diff
@@ -135,10 +136,14 @@ int main(int a_argc, char **a_argv) {
 
       if (gen_keys) {
         string pub_key, priv_key;
+        unsigned char token_key[32];
```
t-ramz, line 139: Magic like Vince Staples

core/server/main.cpp
```diff
@@ -135,10 +136,14 @@ int main(int a_argc, char **a_argv) {
...
+
         generateKeys(pub_key, priv_key);
+        CipherEngine::generateEncryptionKey(token_key);
+
 
-        string fname = config.cred_dir + "datafed-core-key.pub";
-        ofstream outf(fname.c_str());
+       string fname = config.cred_dir + "datafed-core-key.pub";
+       ofstream  outf(fname.c_str());
```
t-ramz, line 146: Whitespace

t-ramz (commented):
Will actually need another go for some of the C++ testing, but here's some more comments

common/include/common/CipherEngine.hpp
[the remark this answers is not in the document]
JoshuaSBrown: Ah, I see, looks like this hasn't been used consistently in the project yet.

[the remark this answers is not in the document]
JoshuaSBrown:
```
/**
 *
 **/
 ```
 
 This is the typical format for Doxygen to pick up the comments and auto generate documentation.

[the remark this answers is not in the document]
Comment (JoshuaSBrown) [reply common/include/common/CipherEngine.hpp:10]: As an example the comments might look something like this

```
/**
 * @class CipherEngine
 * @brief Provides symmetric encryption and decryption functionalities using a 256-bit key.
 **/
class CipherEngine
{
private:
    /**
     * @brief The 256-bit encryption key.
     **/
    unsigned char key[32];

    /**
     * @brief Internal utility to handle encryption/decryption errors.
     **/
    static void handleErrors(void);

public:
    /**
     * @brief Generates a new random 256-bit (32-byte) encryption key.
     * @param token_key A 32-byte buffer to be filled with the generated key.
     **/
    static void generateEncryptionKey(unsigned char token_key[32]);

    /**
     * @brief Generates a new random 128-bit (16-byte) Initialization Vector (IV).
     * @param iv A 16-byte buffer to be filled with the generated IV.
     **/
    static void generateIV(unsigned char iv[16]);
```

File: common/source/CipherEngine.cpp [hunk 27]
```diff
@@ -0,0 +1,227 @@
...
+    CipherEngine::CipherString CipherEngine::createCipherString()
+    {
+        CipherString cs;
+        return cs;
+    }
+    
+    std::unique_ptr<char[]> encode64(const unsigned char* input, int length)
+    {
+        const int pl = 4*((length+2)/3);
+        auto output = std::make_unique<char[]>(pl+1);
```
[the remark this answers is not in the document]
JoshuaSBrown: Yes, that is accurate, I think this is a good use of auto here. It reduces the verbosity of the code without obfuscating the underlyiong types. You also see where people over use auto and essentially hide the underlying type, which I am not a fan of.

common/tests/unit/test_CipherEngine.cpp
```diff
@@ -0,0 +1,148 @@
...
+    readFile("../../build/core/server/datafed-token-key.txt", 32, key);
+    
+    //Construct
```
JoshuaSBrown, line 30: This comment serves no purpose - remove or add more information.

common/tests/unit/test_CipherEngine.cpp
```diff
@@ -0,0 +1,148 @@
...
+    CipherEngine encryptCipher(key);
+
+    //Sets struct CipherString: which contains cipherText, cipherIV, cipherPaddedLen
+    string msg = "1234567890yoDa56Bx5yobvJYEjdGr2YpGYJybE7x4Bq42pQ3zuXCb8YQyn0EqEB7vjPx3GlNlKwkEsMn1234567890";
+
+    //string msg = "Hello World";
+    //CipherEngine::generateIV(iv);
+
+    //Here if parties would like to use their own IV GENERATOR
+    //CipherEngine::CipherString returnObj = testCipher.encrypt(iv, msg);
```
JoshuaSBrown, line 40: This should be a separate test case not a comment.

common/tests/unit/test_CipherEngine.cpp
```diff
@@ -0,0 +1,148 @@
...
+    
+    //Construct
+    CipherEngine encryptCipher(key);
+
+    //Sets struct CipherString: which contains cipherText, cipherIV, cipherPaddedLen
+    string msg = "1234567890yoDa56Bx5yobvJYEjdGr2YpGYJybE7x4Bq42pQ3zuXCb8YQyn0EqEB7vjPx3GlNlKwkEsMn1234567890";
+
+    //string msg = "Hello World";
+    //CipherEngine::generateIV(iv);
+
```
JoshuaSBrown, line 38: Remove this

common/tests/unit/test_CipherEngine.cpp
```diff
@@ -0,0 +1,148 @@
...
+    std::cout << "THIS IS A NEW FLAG" << std::endl;
+    unsigned char key[32];
+    //CipherEngine::generateEncryptionKey(key);
+    readFile("../../build/core/server/datafed-token-key.txt", 32, key);
+    
+    //Construct
+    CipherEngine encryptCipher(key);
+
+    //Sets struct CipherString: which contains cipherText, cipherIV, cipherPaddedLen
+    string msg = "1234567890yoDa56Bx5yobvJYEjdGr2YpGYJybE7x4Bq42pQ3zuXCb8YQyn0EqEB7vjPx3GlNlKwkEsMn1234567890";
```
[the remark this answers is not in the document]
JoshuaSBrown: I think it can also be made const.

common/tests/unit/test_CipherEngine.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+    //string msg = "Hello World";
+    //CipherEngine::generateIV(iv);
+
+    //Here if parties would like to use their own IV GENERATOR
+    //CipherEngine::CipherString returnObj = testCipher.encrypt(iv, msg);
+    
+    CipherEngine::CipherString returnObj2 = encryptCipher.encrypt(msg);
+
```
JoshuaSBrown, line 43: change name 'returnObj2' is not a great choice. Maybe something like encoded_encrypted_packet or something like that.

common/tests/unit/test_CipherEngine.cpp
```diff
@@ -0,0 +1,148 @@
...
+    
+    CipherEngine::CipherString returnObj2 = encryptCipher.encrypt(msg);
+ 
+/*
+    std::cout << "Rough Test1:" << encrypted_access_string << std::endl; 
+    std::string testString;
+
+    std::memcpy(returnObj2.encrypted_msg, testString.c_str(), returnObj2.encrypted_msg_len);
+    std::cout << "Rough Test2:" << testString << std::endl;    
+*/
```
JoshuaSBrown, line 50: Remove this.

common/tests/unit/test_CipherEngine.cpp
```diff
@@ -0,0 +1,148 @@
...
+    std::cout << "Encrypted Message:\n" << returnObj2.encrypted_msg.get() << std::endl;
+    std::cout << "IV:\n" << returnObj2.iv.get() << std::endl;    
+    std::cout << "Encrypted Message Len:\n" << returnObj2.encrypted_msg_len << std::endl;
+
+
+    //WRITE CODE TO MOCK PUTTING IT IN A DB
+
+
+
+
```
JoshuaSBrown, line 60: Remove extra space.

common/tests/unit/test_CipherEngine.cpp
```diff
@@ -0,0 +1,148 @@
...
+    CipherEngine decryptCipher(key);
+
+    //START OF ENCRYPTION
+    std::string unencrypted_msg;
+    unencrypted_msg = decryptCipher.decrypt(returnObj2);
```
JoshuaSBrown, line 65:
```suggestion
    std::string unencrypted_msg = decryptCipher.decrypt(returnObj2); 
```

common/tests/unit/test_CipherEngine.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+
+
+
+    CipherEngine decryptCipher(key);
+
+    //START OF ENCRYPTION
+    std::string unencrypted_msg;
+    unencrypted_msg = decryptCipher.decrypt(returnObj2); 
+    std::cout << "Unencrypted Message:" << unencrypted_msg << std::endl;
```
JoshuaSBrown, line 66: Again print expected and actual values side by side.

scripts/copy_dependency.sh
```diff
@@ -77,14 +83,22 @@ if [ "$direction" == "from" ]; then
   source_filename="$library_location/$library_name"
   destination_filename="$LIBRARIES_BASE_PATH/$library_name"
   if [[ "$library_name" == *".so" ]]; then
-    source_filename="$source_filename.$library_version"
+    if [ "${library_version}" == "" ]; then
```
t-ramz, line 86: Ok so it looks like the version being empty in the other file was intentional. Maybe we could have a more clear implementation/representation like `"${library_version}" == "base"`

core/server/tests/unit/test_TaskWorker.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+BOOST_AUTO_TEST_CASE(testing_unencrypted_token)
+{
+    SDMS::LogContext log_context;
+    bool needs_update = false;
+
+    //setting up test params
+    libjson::Value test_params;
+
+    test_params.fromString("{\"acc_tok\":\"V4X1XamiOcgmJQQYJn6f00UNFxiPFVsltndKamXuQ15HkbcGAXS1EXReSCKcU+7IubXcblajGRKsvsmBIeg2npLbNFCjRukffbgpfIvXW1ofHRORLhs7JVKRSh/SA1uI\",\"acc_tok_iv\":\"BU6pwgToglmdI4W/eIVaiQ==\",\"acc_tok_len\":96}");
```
t-ramz, line 45: Like Joshua said in another file, we should make sure these values are explained somewhere.

common/tests/unit/test_CipherEngine.cpp
```diff
@@ -0,0 +1,148 @@
...
+    BOOST_CHECK(msg.compare(unencrypted_msg) == 0);
+}
+
+
+BOOST_AUTO_TEST_CASE(test_EncryptionDecryptionJSONValue)
+{
+
+    std::cout << "THIS IS A NEW FLAG" << std::endl;
```
t-ramz, line 74: Should probably clean up any `cout` calls

common/tests/unit/test_CipherEngine.cpp
```diff
@@ -0,0 +1,148 @@
...
+    //Sets struct CipherString: which contains cipherText, cipherIV, cipherPaddedLen
+    string msg = "1234567890yoDa56Bx5yobvJYEjdGr2YpGYJybE7x4Bq42pQ3zuXCb8YQyn0EqEB7vjPx3GlNlKwkEsMn1234567890";
+
+    //string msg = "Hello World";
+    //CipherEngine::generateIV(iv);
+
+    //Here if parties would like to use their own IV GENERATOR
+    //CipherEngine::CipherString returnObj = testCipher.encrypt(iv, msg);
+    
+    CipherEngine::CipherString returnObj2 = testCipher.encrypt(msg);
```
t-ramz, line 91: naming

common/tests/unit/test_CipherEngine.cpp
```diff
@@ -0,0 +1,148 @@
...
+    //
+    //
+    CipherEngine testCipher2(key);
+    CipherEngine::CipherString encoded_access_obj = testCipher2.createCipherString();
+    encoded_access_obj.encrypted_msg_len = obj.getNumber("access_len");
+    encoded_access_obj.encrypted_msg = std::make_unique<char[]>(129); // add 1 for null terminator
+
+    std::string access = obj.getString("access");
+    memcpy(encoded_access_obj.encrypted_msg.get(), access.c_str(), 128);
+    encoded_access_obj.encrypted_msg[128] = '\0'; // null terminate
```
t-ramz, line 112: Magic number

common/tests/unit/test_CipherEngine.cpp
```diff
@@ -0,0 +1,148 @@
...
+  
+  // Do the same for IV
+    std::string access_iv = obj.getString("access_iv");
+    encoded_access_obj.iv = std::make_unique<char[]>(25);
+    memcpy(encoded_access_obj.iv.get(), access_iv.c_str(), 24);
+    encoded_access_obj.iv[24] = '\0';
```
t-ramz, line 118: Magic number, but this time it's 24

common/tests/unit/test_CipherEngine.cpp
```diff
@@ -0,0 +1,148 @@
...
+    encoded_access_obj.encrypted_msg_len = obj.getNumber("access_len");
+    encoded_access_obj.encrypted_msg = std::make_unique<char[]>(129); // add 1 for null terminator
+
+    std::string access = obj.getString("access");
+    memcpy(encoded_access_obj.encrypted_msg.get(), access.c_str(), 128);
+    encoded_access_obj.encrypted_msg[128] = '\0'; // null terminate
+  
+  // Do the same for IV
+    std::string access_iv = obj.getString("access_iv");
+    encoded_access_obj.iv = std::make_unique<char[]>(25);
```
t-ramz, line 116: This should be magic number + 1, and specify the null term consideration

common/include/common/CipherEngine.hpp
```diff
@@ -0,0 +1,49 @@
+#ifndef CIPHER_ENGINE_HPP
+#define CIPHER_ENGINE_HPP
+#pragma once
```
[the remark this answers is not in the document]
JoshuaSBrown: If we weren't using C++17, probably would be, and redundant to what? The header guards? Yes. This all depends on what version of a compiler you are using, it would probably be fine to take the header guards off and just rely on the pragma once statement, but I haven't tested this and keeping the header guards is not hurting anything.

scripts/install_core_dependencies.sh
```diff
@@ -10,7 +10,7 @@ PROJECT_ROOT=$(realpath ${SOURCE}/..)
 source "${PROJECT_ROOT}/scripts/utils.sh"
 source "${PROJECT_ROOT}/scripts/dependency_install_functions.sh"
 
-packages=("libtool" "build-essential" "g++" "gcc" "make" "libboost-all-dev" "pkg-config" "autoconf" "automake" "unzip" "wget" "rapidjson-dev" "libkrb5-dev" "git" "python3-pkg-resources" "python3-pip" "python3-venv")
+packages=("libtool" "build-essential" "g++" "gcc" "make" "libboost-all-dev" "pkg-config" "autoconf" "automake" "unzip" "wget" "rapidjson-dev" "libkrb5-dev" "git" "python3-pkg-resources" "python3-pip" "python3-venv" "gcc-multilib")
```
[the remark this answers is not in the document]
JoshuaSBrown: Ya, I'm not sure we need it, if we do we need to have a clear comment of what is using it and why, I think it had something to do with trying to get libsodium to compile with 64 bit, but I think the probably ended up being due to something else. @megatnt1122 try removing it and see if everything builds correctly.

common/source/CipherEngine.cpp
```diff
@@ -0,0 +1,227 @@
...
+        abort();
+    }
+    
+    CipherEngine::CipherString CipherEngine::createCipherString()
+    {
+        CipherString cs;
+        return cs;
+    }
+    
+    std::unique_ptr<char[]> encode64(const unsigned char* input, int length)
```
AronPerez, line 33: We can enhance memory management by utilizing `std::vector`, as it efficiently manages the output buffer through automatic memory allocation and deallocation.
```c++
std::vector<char> encode64(const unsigned char* input, int length) { 
            ...                                                         
          return output;                                  
}         
```

common/source/CipherEngine.cpp
AronPerez, line 44: std::vector<unsigned char> for managing the output buffer.

```c++
std::vector<unsigned char> decode64(const char *input, int length) {
// ... 
return output;
}
```

common/include/common/CipherEngine.hpp
```diff
@@ -0,0 +1,49 @@
...
+            unsigned char encrypted_msg[128];
+            unsigned char iv[16];
+            int encrypted_msg_len;
+        };
+
+        struct CipherString
+        {
+            std::unique_ptr<char[]> encrypted_msg;
+            std::unique_ptr<char[]> iv;
+            int encrypted_msg_len;
```
AronPerez, line 35: I think we can use `std::vector<char>` for `encrypted_msg` and `iv` to simplify memory management.

AronPerez (changes requested):
Looking good, lots of good feedback above. I'm just focusing on error handling and memory management

common/include/common/CipherEngine.hpp
```diff
@@ -0,0 +1,49 @@
...
+        };
```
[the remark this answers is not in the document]
t-ramz: Agreed on this. It would also make things easier to document if we have versions

core/database/foxx/api/lib/user_token.js
```diff
@@ -128,6 +128,11 @@ class UserToken {
             expires_in: 0,
             token_type: g_lib.AccessTokenType.ACCESS_SENTINEL,
             scopes: "",
+            access_iv: "",
+            access_len: 0,
+            refresh_iv: "",
```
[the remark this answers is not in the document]
t-ramz: IV is the initialization [value|vector] https://en.m.wikipedia.org/wiki/Initialization_vector

Basically it's a second part of the cryptographic key

common/source/CipherEngine.cpp
t-ramz: Are `std::unique_ptr<char[]>` workable as drop-in replacements API-wise? If not, I think we should leave the code as-is for now since it's memory managed already. A move to Vector would be better as a follow-up IMO.

JoshuaSBrown:
Ok now we are also seeing this:

```
2025-07-03T19:58:30.295217Z ERROR /datafed/source/core/server/Config.cpp:loadRepositoryConfig:47 { "thread_name": "core_server-repoCacheThread", "thread_id": "2", "message": "Ignoring repo/datafed-at-hot-potato - invalid public key: " }
```

Looks like the zeroMQ keys are behaving oddly.

JoshuaSBrown:
It looks like that failed because curl was missing from python client machine which was running the tests.

```
Warning: Found 2 installation flag files with prefix '.nvm_installed-'. Cleaning up...
Removed all existing installation flag files with prefix '.nvm_installed-'.
/home/gitlab-runner/builds/gYxDkX87B/0/dlsw/datafed/datafed/scripts/dependency_install_functions.sh: line 430: /shared/install/bin/curl: No such file or directory
```

JoshuaSBrown:
@nedvedba can you take a look at this and give some input.

```
58.72 Setting up libboost-mpi-python-dev (1.74.0.3) ...
58.72 Setting up libboost-mpi-dev (1.74.0.3) ...
58.73 Setting up libboost-all-dev (1.74.0.3) ...
58.74 Processing triggers for libc-bin (2.36-9+deb12u10) ...
58.90 DEPENDENCIES (setuptools setuptools sphinx sphinx-rtd-theme sphinx-autoapi)
58.90 /datafed/source/scripts/dependency_install_functions.sh: line 159: python3.9: command not found
------
Dockerfile.dependencies:51
--------------------
  49 |     # Core dependencies
  50 |     COPY ./scripts/install_core_dependencies.sh ${BUILD_DIR}/scripts/
  51 | >>> RUN DEBIAN_FRONTEND=noninteractive TZ=Etc/UTC ${BUILD_DIR}/scripts/install_dependencies.sh -a -r -z -w
  52 |     
  53 |     # Web dependencies
--------------------
ERROR: failed to solve: process "/bin/bash -c DEBIAN_FRONTEND=noninteractive TZ=Etc/UTC ${BUILD_DIR}/scripts/install_dependencies.sh -a -r -z -w" did not complete successfully: exit code: 127
```

I'm not sure why the python interpreter is not being found here. 

https://code.ornl.gov/dlsw/datafed/datafed/-/jobs/3353972

common/source/Util.cpp
```diff
@@ -74,6 +96,10 @@ void generateKeys(std::string &a_pub_key, std::string &a_priv_key) {
   a_priv_key = secret_key;
 }
 
+void generateKeys(std::string &a_pub_key, std::string &a_priv_key)
+{
+    generateZMQKeys(a_pub_key, a_priv_key);
+}
```
JoshuaSBrown, line 102: ?

core/docker/Dockerfile
```diff
@@ -45,7 +45,8 @@ RUN ${BUILD_DIR}/scripts/generate_datafed.sh && \
 		-DBUILD_WEB_SERVER=False		\
 		-DBUILD_DOCS=False					\
 		-DBUILD_PYTHON_CLIENT=False	\
-		-DBUILD_FOXX=False
+		-DBUILD_FOXX=False \
+        -DCMAKE_BUILD_TYPE=Debug
```
JoshuaSBrown, line 49: I don't think you meant to commit this. We don't want the images built with Debug code. That was probably during testing.

core/server/AuthMap.cpp
```diff
@@ -2,7 +2,7 @@
 // Local private includes
 #include "AuthMap.hpp"
 #include "DatabaseAPI.hpp"
-
+#include "Config.hpp"
```
JoshuaSBrown, line 5: Is this header actually being used in this file?

common/include/common/CipherEngine.hpp
```diff
@@ -0,0 +1,156 @@
...
+
+//Local include
+#include <string>
+#include <memory>
+
+namespace SDMS{
+
+/**
+ * @class CipherEngine
+ * @brief Provides symmetric encryption and decryption functoinalities using a 256-bit key.
```
JoshuaSBrown, line 16:
```suggestion
 * @brief Provides symmetric encryption and decryption functionalities using a 256-bit key.
```

common/source/Util.cpp
```diff
@@ -63,7 +85,7 @@ size_t curlBodyReadCB(char *ptr, size_t size, size_t nmemb, void *userdata) {
   return len;
 }
 
-void generateKeys(std::string &a_pub_key, std::string &a_priv_key) {
+void generateZMQKeys(std::string &a_pub_key, std::string &a_priv_key) {
```
JoshuaSBrown, line 88:
```suggestion
void generateKeys(std::string &a_pub_key, std::string &a_priv_key) {
```

core/database/foxx/api/tasks.js
```diff
@@ -1307,9 +1319,13 @@ var tasks_func = (function () {
...
-                        acc_tok: tokens.acc_tok,
-                        ref_tok: tokens.ref_tok,
-                        acc_tok_exp_in: tokens.acc_tok_exp_in,
+                        access_token: tokens.access_token,
+                        refresh_token: tokens.refresh_token,
+                        access_exp_in: tokens.access_exp_in,
+                        access_iv: tokens.access_iv,
+                        access_len: tokens.access_len,
+                        refresh_iv: tokens.refresh_iv,
+                        refresh_len: tokens.refresh_len,
```
JoshuaSBrown, line 1328: You can use spread syntax here.
```suggestion
                        ...tokens.access_token
```

core/server/DatabaseAPI.cpp
```diff
@@ -331,10 +335,12 @@ void DatabaseAPI::userGetAccessToken(
     std::string &a_acc_tok, std::string &a_ref_tok, uint32_t &a_expires_in,
     const std::string collection_id, const std::string collection_type,
     bool &needs_consent, int &token_type, // TODO: use underlying type?
-    std::string &scopes, LogContext log_context) {
+    std::string &scopes, bool &needs_encrypted, LogContext log_context) {
+
+  DL_DEBUG(log_context, "User Get Access Token");
```
JoshuaSBrown, line 340: Can we get rid of this?

```suggestion
```

core/server/TaskWorker.cpp
```diff
@@ -190,12 +193,87 @@ void TaskWorker::workerThread(LogContext log_context) {
...
+
+        //Prep IV into a char[]
+        string iv_str = obj.getString(token + "_iv");  // assume known size
+        encoded_obj.iv = std::unique_ptr<char[]>(new char[iv_str.size() + 1]);  // +1 for null terminator
+        std::memcpy(encoded_obj.iv.get(), iv_str.c_str(), iv_str.size() + 1);     // copy including '\0'
+
+        //Decrypt it:
+        string decrypted_token = cipher.decrypt(encoded_obj, log_context);
+
+        return decrypted_token;
```
JoshuaSBrown, line 227:
```suggestion
        return cipher.decrypt(encoded_obj, log_context);
```

core/server/tests/unit/test_TaskWorker.cpp
```diff
@@ -0,0 +1,132 @@
...
+                           "\",\"acc_tok_iv\":\"" + std::string(returnObj.iv.get()) +
+                           "\",\"acc_tok_len\":96}");
+
+    const libjson::Value::Object &obj = test_params.asObject();
+
+    // Check whether the token needs to be updated (based on IV, length, etc.)
+    needs_update = TaskWorkerTestAccess::tokenNeedsUpdate(obj);
+
+    // Decrypt or prepare the token using the access function
+    std::string acc_tok = TaskWorkerTestAccess::prepToken(obj, TaskWorker::enumToString(access_token_name), cipher_key_path, needs_update, log_context);
```
JoshuaSBrown, line 100:
```suggestion
    std::string acc_tok = TaskWorkerTestAccess::prepToken(obj, TaskWorker::enumToString(TaskWorker::Token_Name::ACCESS), cipher_key_path, needs_update, log_context);
```

core/server/tests/unit/test_TaskWorker.cpp
```diff
@@ -0,0 +1,132 @@
...
+    BOOST_CHECK(acc_tok == "1234567890yoDa56Bx5yobvJYEjdGr2YpGYJybE7x4Bq42pQ3zuXCb8YQyn0EqEB7vjPx3GlNlKwkEsMn1234567890"
+                 && needs_update == false);
+}
+
+// Create a test case for handling unencrypted tokens
+BOOST_AUTO_TEST_CASE(testing_unencrypted_token)
+{
+    SDMS::LogContext log_context;
+    bool needs_update = false;
+    TaskWorker::Token_Name access_token_name = TaskWorker::Token_Name::ACCESS;
```
JoshuaSBrown, line 112:
```suggestion
```

core/server/tests/unit/test_TaskWorker.cpp
```diff
@@ -0,0 +1,132 @@
...
+                           "\"acc_tok_iv\":\"\",\"acc_tok_len\":0}");
+
+    const libjson::Value::Object &obj = test_params.asObject();
+    std::string cipher_key_path = "";
+
+    // Check whether the unencrypted token should be updated
+    needs_update = TaskWorkerTestAccess::tokenNeedsUpdate(obj);
+
+    // Prepare the token (may involve encrypting it if update is needed)
+    std::string acc_tok = TaskWorkerTestAccess::prepToken(obj, TaskWorker::enumToString(access_token_name), cipher_key_path, needs_update, log_context);
```
JoshuaSBrown, line 126:
```suggestion
    std::string acc_tok = TaskWorkerTestAccess::prepToken(obj, TaskWorker::enumToString(TaskWorker::Token_Name::ACCESS), cipher_key_path, needs_update, log_context);
```

common/source/CipherEngine.cpp
```diff
@@ -0,0 +1,218 @@
...
+        const int paddedLength = SDMS::CipherEngine::BASE64_ENCODED_BLOCK_SIZE*((length + (SDMS::CipherEngine::BASE64_INPUT_BLOCK_SIZE-1))/SDMS::CipherEngine::BASE64_INPUT_BLOCK_SIZE);
+        auto output = std::make_unique<char[]>(paddedLength+SDMS::CipherEngine::NULL_TERMINATOR_SIZE);
+        std::fill_n(output.get(), paddedLength + SDMS::CipherEngine::NULL_TERMINATOR_SIZE, 0);  // manually zero-initialize
+        const int outputLength = EVP_EncodeBlock(reinterpret_cast<unsigned char*>(output.get()), input, length);
+        if (paddedLength != outputLength) 
+        {  DL_ERROR(log_context, "Output Length (" << outputLength <<") and Predicted Padded Length ("<< paddedLength <<" ) of encoded bytes not equal!"); } 
+        return output;
+    }
+
+    std::unique_ptr<unsigned char[]> CipherEngine::decode64(const char* input,const int length, LogContext log_context) {
```
JoshuaSBrown, line 45: This decode and encode methods can be made const because they are not changing the state of the instance.
```suggestion
    std::unique_ptr<unsigned char[]> CipherEngine::decode64(const char* input,const int length, LogContext log_context) const {
```

common/source/CipherEngine.cpp
```diff
@@ -0,0 +1,218 @@
...
+using namespace std;
+namespace SDMS{
+      
+    void CipherEngine::handleErrors(void)
+    {
+        ERR_print_errors_fp(stderr);
+        abort();
+    } 
+    
+    std::unique_ptr<char[]> CipherEngine::encode64(const unsigned char* input,const int length, LogContext log_context)
```
JoshuaSBrown, line 32:
```suggestion
    std::unique_ptr<char[]> CipherEngine::encode64(const unsigned char* input,const int length, LogContext log_context) const
```

common/include/common/CipherEngine.hpp
```diff
@@ -0,0 +1,156 @@
...
+{
+    protected:
+        /**
+         * @brief Base64 encodes an input of a specified length and returns the output
+         * @param input Pointer to input bytes to encode
+         * @param length Number of bytes in input
+         * @param log_context Context for logging
+         * @return Base64 encoded null-terminated char array
+         */
+        std::unique_ptr<char[]> encode64(const unsigned char* input, const int length, LogContext log_context);
```
JoshuaSBrown, line 28:
```suggestion
        std::unique_ptr<char[]> encode64(const unsigned char* input, const int length, LogContext log_context) const;
```

common/source/Util.cpp
```diff
@@ -34,6 +39,21 @@ std::string exec(const char *cmd) {
...
 }
 
+int readFile(std::string fileName, int arraySize, unsigned char* array)
+{
+    //Converting Key for encryption funct
+    unsigned char keyChar[arraySize] = {};
+    //Grabbing key
+    std::ifstream keyFile(fileName, std::ios::binary);
+
+    keyFile.read(reinterpret_cast<char*>(keyChar),arraySize);
```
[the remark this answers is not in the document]
JoshuaSBrown: Aaron makes a good point.

core/database/foxx/api/lib/user_token.js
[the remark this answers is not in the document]
AronPerez: 🫡

common/source/CipherEngine.cpp
```diff
@@ -0,0 +1,218 @@
...
+
+    std::string CipherEngine::decrypt(const CipherString& encoded_encrypted_string, LogContext log_context)
+    {
+    EVP_CIPHER_CTX *ctx = nullptr;
+
+    //converts the cipherstring back to a unsigned char
+    std::unique_ptr<unsigned char[]> ciphertext = decode64(encoded_encrypted_string.encrypted_msg.get(), static_cast<int>(strlen(encoded_encrypted_string.encrypted_msg.get())), log_context);
+    std::unique_ptr<unsigned char[]> iv = decode64(encoded_encrypted_string.iv.get(), static_cast<int>(strlen(encoded_encrypted_string.iv.get())),log_context);
+    /* Create and initialise the context */
+    if(!(ctx = EVP_CIPHER_CTX_new()))
```
AronPerez, line 171: Consider caching `EVP_CIPHER_CTX` for better performance in high-throughput scenarios

common/source/CipherEngine.cpp
```diff
@@ -0,0 +1,218 @@
...
+         * Initialise the encryption operation.
+         * IMPORTANT - ensure you use a key
+         * and IV size appropriate for your cipher
+         * In this example we are using 256 bit AES
+         * (i.e. a 256 bit key). The
+         * IV size for *most* modes is the same
+         * as the block size. For AES this
+         * is 128 bits
+         */
+        if(1 != EVP_EncryptInit_ex(ctx, EVP_aes_256_cbc(), NULL, key, iv))
```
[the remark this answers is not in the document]
AronPerez: - https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-38d.pdf - Recommendation for
  GCM
  - https://robertheaton.com/2013/07/29/padding-oracle-attack/ - Explanation of vulnerability
  - https://wiki.openssl.org/index.php/EVP_Authenticated_Encryption_and_Decryption

common/source/CipherEngine.cpp
```diff
@@ -0,0 +1,218 @@
...
+#include <vector>
+#include <fstream>
+#include <iomanip>
+using namespace std;
+namespace SDMS{
+      
+    void CipherEngine::handleErrors(void)
+    {
+        ERR_print_errors_fp(stderr);
+        abort();
```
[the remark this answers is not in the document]
AronPerez: - https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#e12-use-noexcept-when-exiting-a-function-because-of-a-throw-is-impossible-or-unacceptable - Proper exception handling
  - https://www.openssl.org/docs/man1.1.1/man3/ERR_get_error.html

common/include/common/CipherEngine.hpp
```diff
@@ -0,0 +1,156 @@
...
+        static const int BASE64_ENCODED_BLOCK_SIZE = 4;
+
+        /** Base64 input block size (always 3) */
+        static const int BASE64_INPUT_BLOCK_SIZE = 3;
+
+        /** Null terminator size */
+        static const int NULL_TERMINATOR_SIZE = 1;
+
+        /** Initialization vector length in bytes */
+        static const int IV_LENGTH = 16;
```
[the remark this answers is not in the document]
AronPerez: - https://www.openssl.org/docs/man1.1.1/man3/EVP_CIPHER_iv_length.html
  - https://www.cryptopp.com/wiki/Initialization_Vector

common/source/CipherEngine.cpp
```diff
@@ -0,0 +1,218 @@
...
+        if (RAND_bytes(token_key, SDMS::CipherEngine::KEY_LENGTH) != 1)
+        {
+            handleErrors();
+        }
+    }
+
+
+    CipherEngine::CipherBytes CipherEngine::encryptAlgorithm(unsigned char *iv, const string& msg, LogContext log_context)
+    {
+        EVP_CIPHER_CTX *ctx = nullptr;
```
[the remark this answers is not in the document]
AronPerez: - https://cheatsheetseries.owasp.org/cheatsheets/Input_Validation_Cheat_Sheet.html
  - https://cwe.mitre.org/data/definitions/20.html

common/source/CipherEngine.cpp
```diff
@@ -0,0 +1,218 @@
...
+// Standard includes
+#include <iostream>
+#include <exception>
+#include <memory>
+#include <string.h>
+#include <string>
+#include <vector>
+#include <fstream>
+#include <iomanip>
+using namespace std;
```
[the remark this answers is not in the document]
AronPerez: - https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf7-dont-write-using-namespace-at-global-scope-in-a-header-file
  - https://google.github.io/styleguide/cppguide.html#Namespaces

common/source/CipherEngine.cpp
```diff
@@ -0,0 +1,218 @@
...
+        //setting IV for the resulting obj:
+        for(int i = 0; i < SDMS::CipherEngine::IV_LENGTH; i++)
+        {
+            bytes_result.iv[i] = iv[i];
+        }
+
+       vector<unsigned char> msg_unsigned(msg.begin(), msg.end());
+
+        /* Create and initialise the context */
+        if(!(ctx = EVP_CIPHER_CTX_new()))
```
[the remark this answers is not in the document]
AronPerez: - https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#r1-manage-resources-automatically-using-resource-handles-and-raii-resource-acquisition-is-initialization
- https://www.oreilly.com/library/view/effective-modern-c/9781491908419/

common/source/CipherEngine.cpp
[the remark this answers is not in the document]
AronPerez: - https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#r1-manage-resources-automatically-using-resource-handles-and-raii-resource-acquisition-is-initialization
- https://www.oreilly.com/library/view/effective-modern-c/9781491908419/

common/source/CipherEngine.cpp
```diff
@@ -0,0 +1,218 @@
...
+        ERR_print_errors_fp(stderr);
+        abort();
+    } 
+    
+    std::unique_ptr<char[]> CipherEngine::encode64(const unsigned char* input,const int length, LogContext log_context) const
+    {
+        // Calculate the padded length based on the input length:
+        // (length + 2) / 3 gives the number of 3-byte blocks (rounded up), multiplied by 4 gives the number of base64 characters required.
+        const int paddedLength = SDMS::CipherEngine::BASE64_ENCODED_BLOCK_SIZE*((length + (SDMS::CipherEngine::BASE64_INPUT_BLOCK_SIZE-1))/SDMS::CipherEngine::BASE64_INPUT_BLOCK_SIZE);
+        auto output = std::make_unique<char[]>(paddedLength+SDMS::CipherEngine::NULL_TERMINATOR_SIZE);
```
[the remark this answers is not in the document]
AronPerez: - https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique
  - https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#e6-use-raii-to-prevent-leaks

common/include/common/CipherEngine.hpp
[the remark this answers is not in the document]
megatnt1122: I was looking at that and hardcodeed the IV length because C++ needs to know array sizes before running the program right? but EVP_CIPHER_iv_length() only gives that info at runtime which would error out I believe.

core/server/CoreServer.cpp
```diff
@@ -138,8 +139,8 @@ void Server::waitForDB() {
 
   for (int i = 0; i < 10; i++) {
     try {
-      DatabaseAPI db_client(m_config.db_url, m_config.db_user,
-                            m_config.db_pass);
+      std::cout << "waitForDB thread, cred_dir " << m_config.cred_dir << std::endl;
```
JoshuaSBrown, line 142:
```suggestion
```

core/server/TaskMgr.cpp
```diff
@@ -80,7 +80,8 @@ void TaskMgr::initialize(LogContext log_context) {
   // Load ready & running tasks and schedule workers
   // TODO This will break if there are too many tasks - must implement a paging
   // system to load chunks of tasks.
-  DatabaseAPI db(m_config.db_url, m_config.db_user, m_config.db_pass);
+  std::cout << "Task manager initialize creating db, cred dir is: " << m_config.cred_dir << std::endl;
```
JoshuaSBrown, line 83:
```suggestion
```

core/server/TaskMgr.cpp
```diff
@@ -218,7 +219,8 @@ void TaskMgr::maintenanceThread(LogContext log_context, int thread_id) {
 
 void TaskMgr::purgeTaskHistory(LogContext log_context) const {
   try {
-    DatabaseAPI db(m_config.db_url, m_config.db_user, m_config.db_pass);
+    std::cout << "Task manager purgeTaskHistory creating db, cred dir is: " << m_config.cred_dir << std::endl;
```
JoshuaSBrown, line 222:
```suggestion
```

core/server/TaskWorker.cpp
```diff
@@ -190,12 +193,93 @@ void TaskWorker::workerThread(LogContext log_context) {
   } // End of outer while loop
 }
 
+
+//Checks if the tokens are encrypted, if not then it returns the token, if it is encrypted it unencrypts it and returns the unencrypted token
+std::string
+TaskWorker::prepToken(const Value::Object &obj,std::string token, const std::string& cipher_key_path,bool needs_update, LogContext log_context)
+{
+    std::cout << "WHAT THE HELL AM I DOINNNNN HEREEEE" << std::endl;
```
JoshuaSBrown, line 201:
```suggestion
```

core/server/TaskWorker.cpp
```diff
@@ -190,12 +193,93 @@ void TaskWorker::workerThread(LogContext log_context) {
...
+    //if the token's encryption already exists
+    if(!needs_update)
+    {
+        std::cout << "THIS IS A ROBBERY:" + needs_update << std::endl;
```
JoshuaSBrown, line 205:
```suggestion
```

core/server/TaskWorker.cpp
```diff
@@ -190,12 +193,93 @@ void TaskWorker::workerThread(LogContext log_context) {
...
+  Token_Name access_token_name = Token_Name::ACCESS;
+  Token_Name refresh_token_name = Token_Name::REFRESH;
+  bool needs_update = false;
   const Value::Object &obj = a_task_params.asObject();
 
+  std::cout << "WE ENTEREING THE THUNDER DOME" << std::endl;
+  //TokenPrepFuncs
+  needs_update = CipherEngine::tokenNeedsUpdate(obj);
+
+  std::cout << "Needs Update in Task Mana: " + needs_update << std::endl;
```
JoshuaSBrown, line 277:
```suggestion
```

core/server/TaskWorker.cpp
```diff
@@ -190,12 +193,93 @@ void TaskWorker::workerThread(LogContext log_context) {
...
+  //Update the tokens to be unencrypted
+  std::cout << "Calling prepToken from cmdRawDataTransfer cipher key file path is " << me.m_db.cipher_key_file_path << std::endl;
```
JoshuaSBrown, line 279:
```suggestion
```

core/server/AuthMap.cpp
```diff
@@ -197,7 +197,8 @@ bool AuthMap::hasKey(const PublicKeyType pub_key_type,
       return true;
 
     // Check to see if it is a user key
-    DatabaseAPI db(m_db_url, m_db_user, m_db_pass);
+    std::cout << "AuthMap haskey, cred_dir is: " << m_cred_dir << std::endl;
```
JoshuaSBrown, line 200:
```suggestion
```

core/server/AuthMap.cpp
```diff
@@ -234,7 +235,8 @@ std::string AuthMap::getUID(const PublicKeyType pub_key_type,
     }
 
     // It must be a persistent user key
-    DatabaseAPI db(m_db_url, m_db_user, m_db_pass);
+    std::cout << "AuthMap getUID, cred_dir is: " << m_cred_dir << std::endl;
```
JoshuaSBrown, line 238:
```suggestion
```

PR #1489, opened 2025-06-11
[TASK] Remove icons from labels

AronPerez:
## Ticket  

N/A

## Description 

On `generateNoteSpan`, set false for `codes` parameter

 What this fixes:
  - The codes=false parameter makes generateNoteSpan() return proper HTML <span> elements with CSS classes instead of raw HTML entity codes
  - SVG text elements can properly render these HTML span elements, displaying the icons correctly
  - Icons like error flags (🏴), warnings (⚠️), and info icons (ℹ️) will now appear properly in node labels instead of showing as "&#xe665;" or "&#xe65f;"

## How Has This Been Tested?  

1. Create 2 data records (or 1)
2. Add relationship (doesn't matter)
3. Set annotation on data record
4. Go into provenance

### Before

image

### After

https://github.com/user-attachments/assets/fda33e7d-73be-4f59-bd68-b396e8533687


JoshuaSBrown:
These icons do need to be added back in at some point, to have feature parity with what was there before. Being able to track error propagation visually is nice, but this is not a high priority.

JoshuaSBrown:
Is this description accurate?

> Icons like error flags (🏴), warnings (⚠️), and info icons (ℹ️) will now appear properly in node labels instead of showing as "" or ""

Issue #1529, opened 2025-07-24
[Development, Setup] Update compose setup script


nedvedba:
## Description 

The `compose_generate_env.sh` does not have the ability to use an existing `.env` file or generate certs without needing to generate a new `.env` file which can cause problems with the setup for the compose environment.

## Acceptance Criteria

The generate env script is able to be run to only generate certs using the current `.env` file or have normal functionality OR a separate script is created to split the certificate generation out of the generate environment script.

PR #1545, opened 2025-07-29
[DAPS-1529] Rework generate env script

nedvedba:
## Ticket  

#1529  

## Description 

In this PR I rewrote a lot of the `compose_generate_env.sh` script so its behavior is more consistent and so that it is easier to use an existing `.env` file, and so that you can run it to generate certs even if they already exist.

## How Has This Been Tested?  

I verified multiple times that the files it generates are the same as it was before, and with all combinations of flags.

## Tasks

* [X] - A description of the PR has been provided, and a diagram included if it is a new feature.
* [X] - Labels have been assigned to the pr
* [X] - A user has been assigned to work on the pr

### DIFF
--- a/scripts/compose_generate_env.sh
+++ b/scripts/compose_generate_env.sh
@@ -7,19 +7,23 @@ Help()
 {
   echo "$(basename $0) Build .env file for compose."
   echo
-  echo "Syntax: $(basename $0) [-h|d|r|m]"
+  echo "Syntax: $(basename $0) [-h|d|r|m|c|f]"
   echo "options:"
   echo "-h, --help                        Print this help message"
   echo "-d, --directory                   Directory where .env will be created."
   echo "-r, --repo-images                 Create .env for just repo services."
   echo "-m, --metadata-images             Create .env for just metadata services"
+  echo "-c, --no-overwrite-certs          Do not overwrite existing certificates"
+  echo "-f, --force                       Force overwrite existing .env file"
 }
 
-VALID_ARGS=$(getopt -o hd:mr --long 'help',directory:,repo-images,metadata-images -- "$@")
+VALID_ARGS=$(getopt -o hd:mrfc --long 'help',directory:,repo-images,metadata-images,force,no-overwrite-certs -- "$@")
 
 BUILD_REPO="TRUE"
 BUILD_METADATA="TRUE"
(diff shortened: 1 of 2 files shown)


AronPerez:
Quick security thought - sourcing the .env file directly could be risky if someone sneaks in malicious code. Right now if someone puts `rm -rf /` in the .env file, it would execute.

Maybe consider parsing it more safely:
```bash
# Instead of: source "$env_file"
# Could do something like:
while IFS='=' read -r key value; do
  if [[ "$key" =~ ^DATAFED_ ]]; then
    export "$key"="$value"
  fi
done < "$env_file"
```

Not a huge deal for internal use, but worth thinking about if this gets used more widely.

AronPerez:
The variable naming is a bit confusing - `get_env_value()` looks for variables like `DATAFED_COMPOSE_DOMAIN` (no prefix) but everywhere else in the script uses `local_DATAFED_COMPOSE_DOMAIN` (with prefix). 

It works fine because the .env file has the unprefixed names, but it took me a minute to figure out what was happening. Maybe add a comment explaining this pattern or pick one approach?

scripts/compose_generate_env.sh
```diff
@@ -54,10 +66,50 @@ then
...
+    echo "Reading existing values from $env_file"
+    # Source the .env file to load existing values
+    set -a  # automatically export all variables
+    source "$env_file"
+    set +a  # turn off automatic export
+  fi
+}
+
+# Function to get value with fallback priority: existing .env -> environment -> default
+get_env_value() {
```
AronPerez, line 82: The variable naming is a bit confusing - `get_env_value()` looks for variables like `DATAFED_COMPOSE_DOMAIN` (no prefix), but everywhere else in the script uses `local_DATAFED_COMPOSE_DOMAIN` (with prefix)

It works fine because the `.env` file has the unprefixed names, but I had to figure out what was happening. Maybe add a comment explaining this pattern, or pick one approach?

nedvedba: That pattern was used just because it is the pattern that already existed and is the one that is used in most of the other scripts in DataFed.

AronPerez (approved):

JoshuaSBrown (approved):

PR #1744, opened 2025-10-29
replace os.path.join with urllib.parse.urljoin

### LINKED ISSUE #1742: [Refactor] Remove use of `os.path.join`
## Description 

`generate_web_server_credentials.py` uses python's `os.path.join` to create the redirect URI. This can have issues with its handling of slashes and should be replaced with `urllib`.

## Resources

[Link to line of code in question](https://github.com/ORNL/DataFed/blob/88d2936a7e7e64854965bb13ab41efd38a99ecf2/scripts/globus/generate_web_server_credentials.py#L43C1-L43C14)

infispiel:
## Description 

Resolves #1742 

## How Has This Been Tested?  

Cleared all existing configuration, created new `.env` file and new globus credentials, ran metadata docker compose and successfully logged in using globus account.

## Summary by Sourcery

Switch to urljoin for constructing redirect URLs in generate_web_server_credentials.py to ensure correct URL assembly.

Bug Fixes:
- Correct redirect URL generation by replacing os.path.join with urllib.parse.urljoin.

Enhancements:
- Add urllib.parse import and update REDIRECT_PATH to use urljoin.

### DIFF
--- a/scripts/globus/generate_web_server_credentials.py
+++ b/scripts/globus/generate_web_server_credentials.py
@@ -5,6 +5,7 @@ from globus_sdk import AuthClient, GroupsClient
 import utils
 
 import os, sys
+import urllib.parse
 
 CLIENT_ID = "f8d0afca-7ac4-4a3c-ac05-f94f5d9afce8"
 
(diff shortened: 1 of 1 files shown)


scripts/globus/generate_web_server_credentials.py
```diff
@@ -40,7 +41,7 @@
 DOMAIN = os.getenv("DATAFED_DOMAIN", default_DOMAIN)
 if len(DOMAIN) == 0:
     DOMAIN = default_DOMAIN
-REDIRECT_PATH = os.path.join("https://", DOMAIN, "ui/authn")
+REDIRECT_PATH = urllib.parse.urljoin(f"https://{DOMAIN}","ui/authn")
```
JoshuaSBrown: Honestly, I think would be simpler and fine. urllib.parse module is not really needed unless you have two variables you are combing and they are parts of path, but this is for the most part hard coded. 

```suggestion
REDIRECT_PATH = f"https://{DOMAIN}/ui/authn"
```

JoshuaSBrown (approved):

